PR #1, opened 2018-07-06
Initial commit

jefferai:
TODO:

Docs (in main repo)


path_login.go
```diff
@@ -0,0 +1,239 @@
...
+		if err != nil {
+			return logical.ErrorResponse(errwrap.Wrapf("error validating signature: {{err}}", err).Error()), nil
+		}
+
+		if err := idToken.Claims(allClaims); err != nil {
+			return logical.ErrorResponse(errwrap.Wrapf("unable to successfully parse all claims from token: {{err}}", err).Error()), nil
+		}
+
+		var jtiVal string
+		if jti, ok := allClaims["jti"]; ok {
```
chrishoffman: My reading of the jti claim is that it is not an id but more or less a nonce.  https://tools.ietf.org/html/rfc7519#section-4.1.7

backend.go
```diff
@@ -0,0 +1,130 @@
...
+		Paths: framework.PathAppend(
+			[]*framework.Path{
+				pathLogin(b),
+				pathConfig(b),
+			},
+		),
+	}
+
+	// Seed the type
+	b.parsedPubKeys.Store(([]interface{})(nil))
```
chrishoffman: Any reason why not to control this with the mutex?  I don't see a high concurrent workload for these values changing and the provider is already protected with the mutex.

path_login.go
```diff
@@ -0,0 +1,239 @@
...
+		return logical.ErrorResponse("could not load configuration"), nil
+	}
+
+	// Here is where things diverge. If it is using OIDC discovery, validate
+	// that way; otherwise validate against the locally configured keys. Once
+	// things are validated, we re-unify the request path when evaluating the
+	// claims.
+	allClaims := map[string]interface{}{}
+	switch {
+	case len(config.ParsedJWTPubKeys) != 0:
```
chrishoffman: Did you look at creating a static key set in the `go-oidc` library used below?  I think it might allow you to consolidate the validation library across these two methods.

path_login.go
```diff
@@ -0,0 +1,239 @@
...
+
+		var jtiVal string
+		if jti, ok := allClaims["jti"]; ok {
+			if jtiStr, ok := jti.(string); ok {
+				jtiVal = jtiStr
+			}
+		}
+
+		switch {
+		case config.BoundIssuer != "" && config.BoundIssuer != idToken.Issuer:
```
chrishoffman: I _think_ the go-oidc jib already checks this.  Also, why is this set at the top level config and the role?

path_login.go
```diff
@@ -0,0 +1,239 @@
...
+			return logical.ErrorResponse("iss claim does not match bound issuer"), nil
+		case role.BoundSubject != "" && role.BoundSubject != idToken.Subject:
+			return logical.ErrorResponse("sub claim does not match bound subject"), nil
+		case len(role.BoundAudiences) != 0 && !strutil.StrListSubset(role.BoundAudiences, idToken.Audience):
+			return logical.ErrorResponse("aud claim does not match any bound audience"), nil
+		case role.BoundID != "" && role.BoundID != jtiVal:
+			return logical.ErrorResponse("jti claim does not match bound id"), nil
+		}
```
chrishoffman: I think it might be cool to have a map of custom claims you want to validate agains in a role.

path_login.go
```diff
@@ -0,0 +1,239 @@
...
+				continue
+			}
+			groupAliases = append(groupAliases, &logical.Alias{
+				Name: group,
+			})
+		}
+	}
+
+	resp := &logical.Response{
+		Auth: &logical.Auth{
```
chrishoffman: I believe you need to add the BoundCIDRs to the auth block.

path_login.go
chrishoffman: I guess my point here is that if the value is never to be reused, it probably has limited value as a bound parameter.

jefferai: Yep, thanks for the pointer. I think you're right. I'll take it out. I had thought it was an ability to scope similar to `sub` but it looks like it is not.

path_login.go
jefferai: Agreed, but I'd rather save that for later than do it in the initial cut.

path_login.go
jefferai: I didn't, but will

path_login.go
chrishoffman: 👍

path_login.go
jefferai: Ah, right.

path_login.go
jefferai: Right, it does. I will take out the check.

What do you mean by "why is this set at the top level config and the role"?

path_login.go
jefferai: Done.

path_login.go
jefferai: Done. Issue is at config level, subject is at role level.

backend.go
jefferai: Done.

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+	// Role binding properties
+	BoundAudiences []string                      `json:"bound_audiences"`
+	BoundSubject   string                        `json:"bound_subject"`
+	BoundCIDRs     []*sockaddr.SockAddrMarshaler `json:"bound_cidrs"`
+	UserClaim      string                        `json:"user_claim"`
+	GroupsClaim    string                        `json:"groups_claim"`
+}
+
+// role takes a storage backend and the name and returns the role's storage
+// entryÍ
```
kalafut: typo

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+				logical.ListOperation: b.pathRoleList,
+			},
+			HelpSynopsis:    strings.TrimSpace(roleHelp["role-list"][0]),
+			HelpDescription: strings.TrimSpace(roleHelp["role-list"][1]),
+		},
+		&framework.Path{
+			Pattern: "role/" + framework.GenericNameRegex("name"),
+			Fields: map[string]*framework.FieldSchema{
+				"name": &framework.FieldSchema{
+					Type:        framework.TypeString,
```
kalafut: Can use `framework.TypeLowerCaseString` and eliminate the `ToLower`s elsewhere in this file.

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+	raw, err := s.Get(ctx, "role/"+strings.ToLower(name))
+	if err != nil {
+		return nil, err
+	}
+	if raw == nil {
+		return nil, nil
+	}
+
+	role := new(jwtRole)
+	if err := json.Unmarshal(raw.Value, role); err != nil {
```
kalafut: `raw.DecodeJSON(role)` is a little tidier, and uses jsonutil.

path_login.go
jefferai: The interface sadly doesn't allow for providing just keys but rather requires you to actually do the full verification, so it's no easier. Will keep this as-is for now.

path_role.go
jefferai: Fixed

path_role.go
jefferai: Oh yeah!

path_role.go
jefferai: Done

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return logical.ListResponse(roles), nil
+}
+
+// pathRoleRead grabs a read lock and reads the options set on the role from the storage
+func (b *jwtAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request, data *framework.FieldData) (*logical.Response, error) {
+	roleName := data.Get("name").(string)
+	if roleName == "" {
```
kalafut: I don't think this check (present in the operations, too), is necessary. "name" is a url parameter so the framework won't even let the request get here if it's missing.

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+		resp = &logical.Response{}
+		resp.AddWarning("max_ttl is greater than the system or backend mount's maximum TTL value; issued tokens' max TTL value will be truncated")
+	}
+
+	// Store the entry.
+	entry, err := logical.StorageEntryJSON("role/"+strings.ToLower(roleName), role)
+	if err != nil {
+		return nil, err
+	}
+	if entry == nil {
```
kalafut: Looking at `StorageEntryJSON`, I don't think this can ever be true.

path_role.go
```diff
@@ -0,0 +1,346 @@
...
+// roleStorageEntry stores all the options that are set on an role
+var roleHelp = map[string][2]string{
+	"role-list": {
+		"Lists all the roles registered with the backend.",
+		"The list will contain the names of the roles.",
+	},
+	"role": {
+		"Register an role with the backend.",
+		`A role is required to authenticate with this backend. The role binds
+		Azure instance metadata with token policies and settings.
```
kalafut: Replace Azure-y text.

jefferai: Oh noes I'm found out

path_login.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		if err := idToken.Claims(allClaims); err != nil {
+			return logical.ErrorResponse(errwrap.Wrapf("unable to successfully parse all claims from token: {{err}}", err).Error()), nil
+		}
+
+		switch {
+		case role.BoundSubject != "" && role.BoundSubject != idToken.Subject:
+			return logical.ErrorResponse("sub claim does not match bound subject"), nil
+		case len(role.BoundAudiences) != 0 && !strutil.StrListSubset(role.BoundAudiences, idToken.Audience):
```
kalafut: The path-help for BoundAudiences is, "Comma-separated list of 'aud' claims that are valid for login", and the error response here is consistent with that. But `StrListSubset` is going to required that all idToken audiences claims are present in role.BoundAudiences. I think that's the desired logic and maybe the message wording ("any") is throwing me off?  Not too familiar with it though.

jefferai: Fixed!

path_role.go
jefferai: Fixed

path_login.go
```diff
@@ -0,0 +1,237 @@
...
+			},
+			GroupAliases: groupAliases,
+			InternalData: map[string]interface{}{
+				"role": roleName,
+			},
+			Metadata: map[string]string{
+				"role": roleName,
+			},
+			LeaseOptions: logical.LeaseOptions{
+				Renewable: true,
```
briankassouf: Were you going to make this non-renewable unless it's periodic?

backend.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+			SealWrapStorage: []string{
+				"config",
+			},
+		},
+		Paths: framework.PathAppend(
+			[]*framework.Path{
+				pathLogin(b),
+				pathConfig(b),
+			},
```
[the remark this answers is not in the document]
jefferai: Fixed.

path_config.go
```diff
@@ -0,0 +1,187 @@
...
+
+	entry, err := s.Get(ctx, "config")
+	if err != nil {
+		return nil, err
+	}
+	if entry == nil {
+		return nil, nil
+	}
+
+	var result jwtConfig
```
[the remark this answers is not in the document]
jefferai: Done.

path_login.go
jefferai: Following the azure/kube backends, I decided to just make it renewable; you can always set the max TTL to the same as the TTL.

path_config.go
```diff
@@ -0,0 +1,187 @@
...
+			tr.TLSClientConfig = &tls.Config{
+				RootCAs: certPool,
+			}
+		}
+		tc := &http.Client{
+			Transport: tr,
+		}
+		oidcCtx := context.WithValue(ctx, oauth2.HTTPClient, tc)
+
+		if _, err := oidc.NewProvider(oidcCtx, config.OIDCIssuerURL); err != nil {
```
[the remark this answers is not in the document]
jefferai: Done!

path_config.go
```diff
@@ -0,0 +1,198 @@
...
+	"golang.org/x/oauth2"
+)
+
+func pathConfig(b *jwtAuthBackend) *framework.Path {
+	return &framework.Path{
+		Pattern: `config`,
+		Fields: map[string]*framework.FieldSchema{
+			"oidc_issuer_url": &framework.FieldSchema{
+				Type:        framework.TypeString,
+				Description: "OIDC issuer URL, without any .well-known component (base path)",
```
briankassouf: Might be worth adding a note here and to the pubkeys description below that only one of them could be set?

path_config.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	entry, err := logical.StorageEntryJSON(configPath, config)
+	if err != nil {
+		return nil, err
+	}
+	if err := req.Storage.Put(ctx, entry); err != nil {
+		return nil, err
+	}
+
```
briankassouf: Can you call `b.reset()` down here to invalidate the existing config?

path_config.go
jefferai: Done

path_config.go
jefferai: Yep, good catch.

README.md
```diff
@@ -1 +1,127 @@
-# vault-plugin-auth-jwt
+# Vault Plugin: JWT Auth Backend [![Build Status](https://travis-ci.org/hashicorp/vault-plugin-auth-jwt.svg?branch=master)](https://travis-ci.org/hashicorp/vault-plugin-auth-jwt)
+
+This is a standalone backend plugin for use with [Hashicorp Vault](https://www.github.com/hashicorp/vault).
+This plugin allows for JWTs (including OIDC tokens) to authenticate with Vault.
```
v6: // , Does this imply that we could log in with Generic SAML or Ping?

jefferai: If your SAML provider will give you a JWT, yes.

PR #13, opened 2018-09-18
Improve context lifetime handling for OIDC provider

briankassouf:


briankassouf:
@jesseeichar That should be fixed by this PR, but please let us know if you are still running into issues

PR #14, opened 2018-10-10
Add support for mapping claims to token metadatas

uepoch:
Fixes: https://github.com/hashicorp/vault/issues/5477

This will let user specify a mapping as kv pairs between their JWT claims and returned token's metadatas.

This will be useful to add additional context from private Claims to existing tokens.
Same could be done with aliases I guess with another field to avoid stepping on each others.

### Disclaimer The name of the field is pretty bad, I wasn't really inspired.. let me know if you want me to change it

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -225,6 +225,16 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 			})
 		}
 	}
+	metadatas := map[string]string{
+		"role": roleName,
+	}
+	if ClaimsMapping := role.ClaimsToMetadatas; ClaimsMapping != nil {
+		for claimMapped, metaKey := range ClaimsMapping {
+			if c, ok := allClaims[claimMapped]; ok && metaKey != "" {
+				metadatas[metaKey] = fmt.Sprintf("%+v", c)
+			}
+		}
+	}
 
 	resp := &logical.Response{
 		Auth: &logical.Auth{
@@ -239,9 +249,7 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 			InternalData: map[string]interface{}{
 				"role": roleName,
 			},
-			Metadata: map[string]string{
-				"role": roleName,
-			},
+			Metadata: metadatas,
 			LeaseOptions: logical.LeaseOptions{
 				Renewable: true,
 				TTL:       role.TTL,
--- a/path_login.go
+++ b/path_login.go
@@ -231,7 +231,7 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 	if ClaimsMapping := role.ClaimsToMetadatas; ClaimsMapping != nil {
 		for claimMapped, metaKey := range ClaimsMapping {
 			if c, ok := allClaims[claimMapped]; ok && metaKey != "" {
-				metadatas[metaKey] = fmt.Sprintf("%+v", c)
+				metadatas[metaKey] = fmt.Sprintf("%v", c)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


uepoch:
Not really sure to understand your idea. You propose to specify the format as a JSON map passed in the role entry ?

I think (Did not tried with NumericDate) Sprintf("%#v") takes the GoStringer interface if present, which should handle most cases of numerical / json parsed types ?

avoidik:
@uepoch could you please rebase your PR against the master?

avoidik:
this PR could improve policy templating with metadata from external systems

uepoch:
@avoidik sorry for delay, was in the forest for last two weeks, will rebase :)

@jefferai as we discussed at Hashiconf, I will update the PR to add an optional user-specified "separator" character in the mapped field to handle nested objects

jefferai:
@uepoch A few weeks later I don't really remember the conversation. Can you give some examples of the behavior in a comment, before writing up the code?

uepoch:
Hello, I believe it's the same as what you did for group_claims_delimiter

Be able to pass a string, which, if provided with crawl the JWT fields nested structures using it as a delimiter in the main "claim" name

jefferai:
I think what I'm not really following is the mapping of map keys to claim keys to sub keys to metadata keys. An example would be very useful.

uepoch:
A final idea of the PR could be:

A role with "claims_to_metadatas" field set as 
`
{
  "delegate_for": "ctx:service:delegator"
}
`
would look in the JWT claims for a key called "ctx:service:delegator", and assign it as the "delegate_for" metadata in the returned token

an optional field in role called "claims_to_metadatas_delimiter_pattern" could be set as "::", and it would search for "delegator" key in a nested map keyed "service" in a top-level map keyed "ctx" and assign its value to "delegate_for" instead

Additionnal stuff:
In the PR, I originally used the claim name as the key of 'claims_to_metadatas', which is, after thinking a bad idea as we only care about uniqueness of the metadata keys, so I changed it in the given example
(And the name of the field is still bad)

jefferai:
@uepoch Is there a reason to put this in token metadata as opposed to identity alias metadata? The latter can then be referenced elsewhere in Vault.

I also wonder if a better idea would be to have the user explicitly specify JSON maps. This would force advanced CLI usage, but this is super advanced anyways. So in your example, you'd get:

`{ "delegate_for": {"ctx": {"service": {"delegator"}}} }` This way there's no delimiter problem at all.

uepoch:
Hey @jefferai 

Initially I was going to include them in the alias, but as the alias is identified after the username (by a rather static claim), and the dynamic nature of some claims's value, I was afraid it would generate a lot of unnecessary volatility and writes/etc on the identity store.

In the example, delegate_for could have 100+ different values on valid JWTs at the same time given a username.
Tho I agree that it could be nice to be able to specify alias-related metadatas
 
Hum, yes I guess passing maps like this would get rid of the delimiter problem, that's a good idea

AFAIK(but didn't tested tbh) it's not possible to do nested maps in CLI ?

jefferai:
> Initially I was going to include them in the alias, but as the alias is identified after the username (by a rather static claim), and the dynamic nature of some claims's value, I was afraid it would generate a lot of unnecessary volatility and writes/etc on the identity store.

I guess it depends on your end goal; if it's for audit logs token metadata will show in cleartext in those; if it's to use elsewhere in Vault, eventually, it'd have to be in alias metadata. It can be _both_ so if there are use cases for both, but you only need one right now, can certainly start with one and the other can be added if/when needed. So if token metadata suits your needs let's leave it at that for now.

> AFAIK(but didn't tested tbh) it's not possible to do nested maps in CLI ?

It's not but you can source JSON files and in general I'd consider this to be quite advanced usage so requiring a little more effort on the part of the user seems fine.

kalafut:
@uepoch We've implemented this functionality part of https://github.com/hashicorp/vault-plugin-auth-jwt/pull/24 which will be in the next beta. Claims to be copied are specified via a JSONPointer, and they'll be added to both the token and alias.

PR #16, opened 2019-01-02
Allow configuration of supported signing algorithms

shupp:
By default, only RS256 is supported.  We use RS512 for our JWT service, so we'd prefer to default to both being available.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -36,7 +36,7 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 			},
 			"jwt_supported_algs": {
 				Type:        framework.TypeCommaStringSlice,
-				Description: `A list of supported encoding algorithms. Defaults to empty, therefore the oidc plugin default, which currently is RS256 only.`,
+				Description: `A list of supported signing algorithms. Defaults to RS256.`,
 			},
 			"bound_issuer": {
 				Type:        framework.TypeString,
--- a/path_config.go
+++ b/path_config.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"net/http"
 
 	"context"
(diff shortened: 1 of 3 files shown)


briankassouf:
Yeah, i'd prefer this to be a configuration setting so we can optionally support any of the algorithms

jefferai:
Agreed on configuration setting.

shupp:
Sounds reasonable to me too, thanks for the responses.  I don't have any experience with the config stuff for vault yet, but I should be able to look into it this week.  If any of you have any pointers, they are welcome :)  Otherwise I'll figure it out.

kalafut:
@shupp For integrating the new config parameter, you should be able to copy the pattern used for [jwt_validation_pubkeys](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_config.go#L184). Add a new parameter to that struct and corresponding support in the CRUD handling functions to manage it. Happy to review WIP.

shupp:
Great, thanks very much @kalafut!!

shupp:
@kalafut ready for another review when available.

kalafut:
@shupp Looks pretty good! My main feedback is that I think we should probably forgo the `default` parameter. This will leaving the existing default set of algorithms unchanged. It also keeps the behavior whereby if you're not explicitly setting that in your config, you get the plugin's default values at runtime.

shupp:
@kalafut please excuse the push parade.  Ready for another review at your leisure.

path_config.go
```diff
@@ -33,6 +34,10 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 				Type:        framework.TypeCommaStringSlice,
 				Description: `A list of PEM-encoded public keys to use to authenticate signatures locally. Cannot be used with "oidc_discovery_url".`,
 			},
+			"jwt_supported_algs": {
+				Type:        framework.TypeCommaStringSlice,
+				Description: `A list of supported encoding algorithms. Defaults to empty, therefore the oidc plugin default, which currently is RS256 only.`,
```
kalafut:
```suggestion
				Description: `A list of supported signing algorithms. Defaults to RS256.`,
```

kalafut:
One minor text change, otherwise LGTM. Any other comments @briankassouf ?

path_config.go
shupp: @kalafut LGTM, let me know if you'd like me to rebase and commit this text change, or if you prefer to just squash it on merge.

kalafut: You should have been given a button to just accept the change in the GH UI which will automagically make a new commit. Don't rebase or squash. We'll do that at the end.

PR #19, opened 2019-01-29
JWT metadata update from claims

kalafut:
This PR also removes the groupClaimsDelimiter approach to specifying
the location of nested claims data. JSONPointer is used instead.


tyrannosaurus-becks:
This code looks good! Should we follow semver in this repo? If so, should we do an initial release of the code before this PR and #20 are merged, and then do a minor release after they're merged?

PR #20, opened 2019-01-29
OIDC auth url support

kalafut:
Note the TODO comment in path_oidc.go. I think that function should be changed to never return error responses other than a couple of specific user error messages.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -29,6 +29,7 @@ type jwtAuthBackend struct {
 	l            sync.RWMutex
 	provider     *oidc.Provider
 	cachedConfig *jwtConfig
+	oidcStates   map[string]*oidcState
 
 	providerCtx       context.Context
 	providerCtxCancel context.CancelFunc
@@ -37,6 +38,7 @@ type jwtAuthBackend struct {
 func backend(c *logical.BackendConfig) *jwtAuthBackend {
 	b := new(jwtAuthBackend)
 	b.providerCtx, b.providerCtxCancel = context.WithCancel(context.Background())
+	b.oidcStates = make(map[string]*oidcState)
 
 	b.Backend = &framework.Backend{
 		AuthRenew:   b.pathLoginRenew,
@@ -46,6 +48,7 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 		PathsSpecial: &logical.Paths{
 			Unauthenticated: []string{
 				"login",
+				"oidc/auth_url",
 			},
 			SealWrapStorage: []string{
 				"config",
@@ -58,9 +61,13 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 				pathRole(b),
 				pathConfig(b),
 			},
+			pathOIDC(b),
 		),
 		Clean: b.cleanup,
 	}
+
+	// Start a periodic cleanup of unused state tokens
+	b.stateGC()
 
 	return b
 }
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -1,0 +1,184 @@
+package jwtauth
+
+import (
+	"context"
+	"time"
+
+	oidc "github.com/coreos/go-oidc"
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/vault/helper/base62"
+	"github.com/hashicorp/vault/helper/strutil"
+	"github.com/hashicorp/vault/logical"
+	"github.com/hashicorp/vault/logical/framework"
+	"golang.org/x/oauth2"
+)
+
+// base62 char length for state and nonce parameters, ~160 bits entropy per rfc6749#section-10.10
+const stateLength = 27
+
+var oidcStateTimeout = 2 * time.Minute
+
+// oidcState is created when an authURL is requested. The state identifier is
+// passed throughout the OAuth process.
+type oidcState struct {
+	rolename    string
+	expiration  time.Time
+	nonce       string
+	redirectURI string
+}
+
+func pathOIDC(b *jwtAuthBackend) []*framework.Path {
+	return []*framework.Path{
+		{
+			Pattern: `oidc/auth_url`,
+			Fields: map[string]*framework.FieldSchema{
+				"role": {
+					Type:        framework.TypeLowerCaseString,
+					Description: "The role to issue an OIDC authorization URL against.",
+				},
+				"redirect_uri": {
+					Type:        framework.TypeString,
+					Description: "The OAuth redirect_uri to use in the authorization URL.",
+				},
+			},
+			Callbacks: map[logical.Operation]framework.OperationFunc{
+				logical.UpdateOperation: b.authURL,
+			},
+		},
+	}
+}
+
+// authURL returns a URL used for redirection to receive an authorization code.
+// This path requires a role name, or that a default_role has been configured.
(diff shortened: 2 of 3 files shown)


backend.go
```diff
@@ -58,10 +61,14 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 				pathRole(b),
 				pathConfig(b),
 			},
+			pathOIDC(b),
 		),
 		Clean: b.cleanup,
 	}
 
+	// Start a periodic cleanup of unused state tokens
+	b.stateGC()
```
tyrannosaurus-becks: I like your naming of this, it's succinct and very clear.

path_oidc.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return resp, nil
+}
+
+// createState make an expiring state object, associated with a random state ID
+// that is passed throughout the OAuth process. A nonce is also included in the
+// auth process, and for simplicity will be identical in length/format as the state ID.
+func (b *jwtAuthBackend) createState(rolename, redirectURI string) (string, string, error) {
+	randstr, err := base62.Random(2 * stateLength)
```
tyrannosaurus-becks: At first I thought this wasn't cryptographically random, but I dug into it and I see that it is.

path_oidc.go
```diff
@@ -0,0 +1,184 @@
...
+
+	delete(b.oidcStates, state)
+
+	return s
+}
+
+// stateGC will start a goroutine which periodically deletes all expired states
+// that were never removed via the normal request process.
+func (b *jwtAuthBackend) stateGC() {
+	go func() {
```
tyrannosaurus-becks: Is there any way to tie this func to something that will cleanly shut it down if the `jwtAuthBackend` is terminated? Like what if it were unmounted, or a request context were cancelled? Ideally there'd be a channel of some sort being passed in.

path_oidc.go
```diff
@@ -0,0 +1,184 @@
...
+		for {
+			now := time.Now()
+			b.l.Lock()
+			for k := range b.oidcStates {
+				if now.After(b.oidcStates[k].expiration) {
+					delete(b.oidcStates, k)
+				}
+			}
+			b.l.Unlock()
+			time.Sleep(1 * time.Minute)
```
tyrannosaurus-becks: Since the sleep is in a loop, might be nice to consider using a ticker that goes off every minute. If it were part of a `select`, it would also be possible to listen to some sort of `shutdown` channel or a call context cancellation channel simultaneously, and to return from the goroutine if they were called.

tyrannosaurus-becks: I bet the `ctx` [here](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/895c729e7b08bb8e65e8bc401de6ce0b918da32c/backend.go#L18) would be usable for detecting a graceful shutdown of the enclosing environment.

tyrannosaurus-becks: Looks like a request context wouldn't be applicable in this case since it's only used at the backend level.

tyrannosaurus-becks: To be clearer, I was imagining this more like this:
```
ticker := yadda
for {
    select {
    case <- ticker.C:
        ....
    case <- ctx.Done():
        ....
    }
}
```

kalafut: I'll look into wiring in context.

kalafut: Just pushed a proposed change that's hooked into the Clean function.

path_oidc.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	redirectURI := d.Get("redirect_uri").(string)
+	if redirectURI == "" {
+		return logical.ErrorResponse("missing redirect_uri"), nil
+	}
+
+	if role != nil && role.RoleType == "oidc" {
```
briankassouf: Do we want to return early if role type != oidc?

path_oidc.go
```diff
@@ -0,0 +1,194 @@
...
+		provider, err := b.getProvider(ctx, config)
+		if err != nil {
+			return nil, errwrap.Wrapf("error getting provider for login operation: {{err}}", err)
+		}
+
+		if len(role.AllowedRedirectURIs) == 0 ||
+			strutil.StrListContains(role.AllowedRedirectURIs, redirectURI) {
```
briankassouf: I think we should mandate the role sets a list of allowed redirects, seems risky to allow people to provide the callback unauthenticated

path_oidc.go
kalafut: Yeah that's a little simpler.

path_oidc.go
```diff
@@ -0,0 +1,194 @@
...
+
+	go func() {
+		ticker := time.NewTicker(1 * time.Minute)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case now := <-ticker.C:
+				b.l.Lock()
+				for k := range b.oidcStates {
```
briankassouf: Would using go-cache be a cleaner solution? It would mean we don't have to iterate over all the states

backend.go
```diff
@@ -71,6 +79,7 @@ func (b *jwtAuthBackend) cleanup(_ context.Context) {
 		b.providerCtxCancel()
 	}
 	b.l.Unlock()
+	b.stateGCCancel <- struct{}{}
```
briankassouf: Can we close instead of write to?

path_oidc.go
```diff
@@ -0,0 +1,194 @@
...
+// createState make an expiring state object, associated with a random state ID
+// that is passed throughout the OAuth process. A nonce is also included in the
+// auth process, and for simplicity will be identical in length/format as the state ID.
+func (b *jwtAuthBackend) createState(rolename, redirectURI string) (string, string, error) {
+	randstr, err := base62.Random(2 * stateLength)
+	if err != nil {
+		return "", "", err
+	}
+
+	stateID, nonce := randstr[0:stateLength], randstr[stateLength:]
```
briankassouf: Do we need to use base62 here? feels like go-uuid might be a faster method

path_oidc.go
kalafut: Couple thoughts. First, that's reasonable. I do wonder if we want them to be able to specify something like "*" to disable any checking. They're at least opting in to that, unlike it being the default behavior now.

Second, I should look into special handling for loopback addresses. Per https://tools.ietf.org/html/rfc8252#section-7.3 the providers shouldn't even consider ports for loopback (but many still do), so I think we can probably safely say if the redirect is a loopback, equality shouldn't include port. This has practical benefits for thing like the CLI helper having a potentially ephemeral port.

path_oidc.go
kalafut: Probably. I think this began really simple and I didn't bother, but at this point just dropping in go-cache might be nicer.

path_oidc.go
kalafut: The base62 charset isn't required. I'll switch to go-uuid, but will just be for random bytes that will be printf'd into hex here, since the RFC calls for 160 bits and not 128.

backend.go
kalafut: This went away with go-cache.

briankassouf: Don't forget to remove this

kalafut: Oh yeah. Done.

PR #21, opened 2019-01-31
Add OIDC callback handler

kalafut:
This PR builds on the other two to complete the auth flow via the callback. A fair amount of the diff is with moving code from the original login function into standalone functions to be shared by the JWT and OIDC paths. The initial version of the code had a single login function regardless of path, but it was a confusing `if`/`switch` mess of shared & unique code.

Tests are forthcoming.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -51,6 +51,9 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 			Unauthenticated: []string{
 				"login",
 				"oidc/auth_url",
+				"oidc/callback",
+				"ui", // TODO: remove when Vault UI is ready
+
 			},
 			SealWrapStorage: []string{
 				"config",
@@ -62,6 +65,7 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 				pathRoleList(b),
 				pathRole(b),
 				pathConfig(b),
+				pathUI(b), // TODO: remove when Vault UI is ready
 			},
 			pathOIDC(b),
 		),
--- a/claims_test.go
+++ b/claims_test.go
@@ -1,39 +1,56 @@
 package jwtauth
 
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/go-test/deep"
 )
 
 func TestGetClaim(t *testing.T) {
-	tests := []struct {
-		allClaims map[string]interface{}
-		claim     string
-		expected  interface{}
-	}{
-		{nil, "data", nil},
-		{map[string]interface{}{"data": "foo"}, "data", "foo"},
-		{map[string]interface{}{"data": "foo"}, "data2", nil},
-		{map[string]interface{}{"data": "foo"}, "/data", "foo"},
-		{
-			map[string]interface{}{"data": map[string]interface{}{
-				"foo": "bar",
-			}}, "/data/foo", "bar",
-		},
-		{
-			map[string]interface{}{"data": map[string]interface{}{
-				"foo": "bar",
-			}}, "/data/foo2", nil,
-		},
+	data := `
+{
+  "a": 42,
+  "b": "bar",
+  "c": {
+    "d": 95,
+    "e": [
+      "dog",
+  	  "cat",
+  	  "bird"
+    ],
+	"f": {
+	  "g": "zebra"
+	}
+  }
+}
+`
+	var claims map[string]interface{}
+	if err := json.Unmarshal([]byte(data), &claims); err != nil {
+		t.Fatal(err)
+	}
 
-		{map[string]interface{}{"data": "foo"}, `\`, nil},
+	tests := []struct {
+		claim string
+		value interface{}
+	}{
+		{"a", float64(42)},
+		{"/a", float64(42)},
+		{"b", "bar"},
+		{"/c/d", float64(95)},
+		{"/c/e/1", "cat"},
+		{"/c/f/g", "zebra"},
+		{"nope", nil},
+		{"/c/f/h", nil},
+		{"", nil},
+		{"\\", nil},
 	}
 
 	for _, test := range tests {
-		actual := getClaim(test.allClaims, test.claim)
-		if diff := deep.Equal(actual, test.expected); diff != nil {
-			t.Fatalf("invalid results for claim '%s': %v", test.claim, diff)
+		v := getClaim(claims, test.claim)
+
+		if diff := deep.Equal(v, test.value); diff != nil {
+			t.Fatal(diff)
 		}
 	}
 }
--- a/path_login.go
+++ b/path_login.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	oidc "github.com/coreos/go-oidc"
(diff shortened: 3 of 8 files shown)


claims_test.go
```diff
@@ -1,39 +1,56 @@
...
 import (
+	"encoding/json"
 	"testing"
 
 	"github.com/go-test/deep"
 )
 
 func TestGetClaim(t *testing.T) {
+	data := `
+{
```
tyrannosaurus-becks: Not 100% sure but I _think_ it'll still parse if you tab it over to be at the nested level for data.

path_oidc.go
```diff
@@ -44,6 +58,107 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
...
+
+	var oauth2Config = oauth2.Config{
+		ClientID:     config.OIDCClientID,
+		ClientSecret: config.OIDCClientSecret,
+		RedirectURL:  state.redirectURI,
+		Endpoint:     provider.Endpoint(),
+		Scopes:       []string{oidc.ScopeOpenID},
+	}
+
+	oauth2Token, err := oauth2Config.Exchange(ctx, d.Raw["code"].(string))
```
tyrannosaurus-becks: What would happen here if the code weren't provided, or were provided but as a different type than as a string?

path_oidc.go
```diff
@@ -44,6 +58,107 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
 	}
 }
 
+func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	state := b.verifyState(d.Raw["state"].(string))
```
tyrannosaurus-becks: What would happen here if the state weren't a string or weren't provided?

path_oidc.go
```diff
@@ -44,6 +58,107 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
...
+		return logical.ErrorResponse(err.Error()), nil
+	}
+
+	// Attempt to fetch information from the /userinfo endpoint and merge it with
+	// the existing claims data. A failure to fetch additional information from this
+	// endpoint will not invalidate the authorization flow.
+	if userinfo, err := provider.UserInfo(ctx, oauth2.StaticTokenSource(oauth2Token)); err == nil {
+		userinfo.Claims(&allClaims)
+	} else {
+		b.Logger().Info("error reading /userinfo endpoint", "error", err)
```
tyrannosaurus-becks: Should this be at the Warn level? I usually put errors I'm swallowing at Warn but I'm not sure if that makes sense here.

kalafut:
@tyrannosaurus-becks Yeah I'll add tests to this PR. Just wanted to let anyone have a jump on the code review sooner.

claims_test.go
kalafut: 👍

path_oidc.go
kalafut: Yeah, probably so. I just upped it to Warn by default. I'll still use Info if the error is that the provider simply doesn't support the `/userinfo` endpoint, which isn't really an error to call attention to but just a trait of the provider. I thought there'd be a simple way to know whether they support it before calling the function, but I'm not seeing it.

path_oidc.go
kalafut: Bad things 😛. Will fix...

kalafut: Refactored to use the Get() method which will ensure we'll at least always have a string.

path_oidc.go
kalafut: Refactored to use the Get() method which will ensure we'll at least always have a string.

claims.go
```diff
@@ -11,14 +11,30 @@ import (
...
+	var val interface{}
+	var err error
+
 	if !strings.HasPrefix(claim, "/") {
-		return allClaims[claim]
+		val = allClaims[claim]
+	} else {
+		val, err = pointerstructure.Get(allClaims, claim)
+		if err != nil {
+			return nil
```
tyrannosaurus-becks: Should there be a logged Warning here showing the error?

backend.go
```diff
@@ -51,6 +51,9 @@ func backend(c *logical.BackendConfig) *jwtAuthBackend {
 			Unauthenticated: []string{
 				"login",
 				"oidc/auth_url",
+				"oidc/callback",
+				"ui", // TODO: remove when Vault UI is ready
+
```
tyrannosaurus-becks: Extra line.

claims_test.go
```diff
@@ -1,39 +1,55 @@
...
+		    "d": 95,
+		    "e": [
+		      "dog",
+		  	  "cat",
+		  	  "bird"
+		    ],
+			"f": {
+			  "g": "zebra"
+			}
+		  }
```
tyrannosaurus-becks: Between the JSON linter and gofmt, the JSON here got a little funky! :-) Not sure if it's worth the effort but if it's easy to get this pretty, it would be nice.

path_login.go
```diff
@@ -55,19 +61,16 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return nil, err
 	}
 	if role == nil {
-		return logical.ErrorResponse("role could not be found"), nil
+		return logical.ErrorResponse("role '%s' could not be found", roleName), nil
```
tyrannosaurus-becks: I just learned from Jeff that we can replace ```'%s'``` with `%q` and it will come out the same!

path_login.go
```diff
@@ -267,6 +198,118 @@ func (b *jwtAuthBackend) pathLoginRenew(ctx context.Context, req *logical.Reques
...
+	}
+
+	if err := idToken.Claims(&allClaims); err != nil {
+		return nil, errwrap.Wrapf("unable to successfully parse all claims from token: {{err}}", err)
+	}
+
+	if role.BoundSubject != "" && role.BoundSubject != idToken.Subject {
+		return nil, errors.New("sub claim does not match bound subject")
+	}
+	if len(role.BoundAudiences) > 0 {
```
tyrannosaurus-becks: Since you're looping over `role.BoundAudiences` below, you can omit this check. If the length is zero, it won't enter the loop, and `found` will remain false.

path_login.go
```diff
@@ -267,6 +198,118 @@ func (b *jwtAuthBackend) pathLoginRenew(ctx context.Context, req *logical.Reques
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, errors.New("aud claim does not match any bound audience")
+		}
+	}
+
+	if len(role.BoundClaims) > 0 {
```
tyrannosaurus-becks: Same here, you could just omit this check and get right on with the looping.

path_login.go
tyrannosaurus-becks: Oh, I see you use `%q` below so there's probably some nuance of this that I don't yet know.

path_login.go
```diff
@@ -267,6 +198,118 @@ func (b *jwtAuthBackend) pathLoginRenew(ctx context.Context, req *logical.Reques
...
+	}
+
+	alias := &logical.Alias{
+		Name:     userName,
+		Metadata: metadata,
+	}
+
+	var groupAliases []*logical.Alias
+
+	if role.GroupsClaim != "" {
```
tyrannosaurus-becks: This is just a stylistic thing and feel free not to do it. I had an old coworker who was all about avoiding nesting if possible, because he thought nested logic could get confusing. In this situation, he would have encouraged me to simply do this here and return right away, merely so I could un-nest the remaining:
```
if role.GroupsClaim == "" { 
    return alias, groupAliases, nil
}
// Now everything else doesn't have to be nested.
```
Totally up to you. I don't really think it's super consequential or that the code is hard to reason about, but thought it might be interesting to bring up and see what you think, or if it even matters.

path_login.go
tyrannosaurus-becks: Belatedly, I see why you did this! It was because they only matter if they were set! Ditto with the above.

claims.go
kalafut: Yeah this function should probably be upgraded to return an error. But IIRC, I couldn't even find a way to make pointerstructure.Get() return an error other than "not found". The compilation errors all came from not starting with '/' which is special-cased before this is even reached. I'll double check though.

path_login.go
kalafut: I've done both ways, usually preferring not too much nested content. I think this code was nested before and I just didn't change it, but it would probably be better if restructured.

PR #23, opened 2019-02-09
Add CLI login support + misc cleanup

kalafut:


cli.go
```diff
@@ -0,0 +1,502 @@
...
+	if !ok {
+		port = defaultPort
+	}
+
+	role := m["role"]
+	if role == "" {
+		return nil, errors.New("a 'role' must be specified")
+	}
+
+	secret, err := fetchAuthURL(c, role, mount, port)
```
tyrannosaurus-becks: Should we name this something else? The naming of this method seems slightly confusing to me because it's returning more than an auth URL, and also because below an auth URL is checked for and there's an error if it doesn't exist, which I would expect to happen in _this_ method given its name.

cli.go
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	authURL := secret.Data["auth_url"].(string)
+	if authURL == "" {
+		return nil, errors.New(fmt.Sprintf("Unable to authorize role %q. Check Vault logs for more information.", role))
+	}
+
+	fmt.Fprintf(os.Stderr, "Complete the login via your OIDC provider. Launching browser to:\n\n    %s\n\n\n", authURL)
+	if err := openURL(authURL); err != nil {
+		fmt.Fprintf(os.Stderr, "Error attempting to automatically open browser: '%s'.\nPlease visit the authorization URL manually.", err)
```
tyrannosaurus-becks: Should we return here? I haven't used this method too much.

cli.go
```diff
@@ -0,0 +1,502 @@
...
+
+		secret, err := c.Logical().ReadWithData(fmt.Sprintf("auth/%s/oidc/callback", mount), data)
+		if err != nil {
+			summary, detail := parseError(err)
+			response = errorHTML(summary, detail)
+		} else {
+			response = successHTML
+		}
+
+		w.Write([]byte(response))
```
tyrannosaurus-becks: Should we also write a status code in the header here? Or does it automatically write a 200 if unspecified?

cli.go
```diff
@@ -0,0 +1,502 @@
...
+type loginResp struct {
+	secret *api.Secret
+	err    error
+}
+
+func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, error) {
+	// handle ctrl-c while waiting for the callback
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
+	defer signal.Stop(ch)
```
tyrannosaurus-becks: Could ch be named something like ctrlC or interrupt or something more descriptive?

cli.go
```diff
@@ -0,0 +1,502 @@
...
+		return nil, errors.New(fmt.Sprintf("Unable to authorize role %q. Check Vault logs for more information.", role))
+	}
+
+	fmt.Fprintf(os.Stderr, "Complete the login via your OIDC provider. Launching browser to:\n\n    %s\n\n\n", authURL)
+	if err := openURL(authURL); err != nil {
+		fmt.Fprintf(os.Stderr, "Error attempting to automatically open browser: '%s'.\nPlease visit the authorization URL manually.", err)
+	}
+
+	// Set up callback handler
+	http.HandleFunc(fmt.Sprintf("/v1/auth/%s/oidc/callback", mount), func(w http.ResponseWriter, req *http.Request) {
```
tyrannosaurus-becks: What would happen if someone tried to auth using a particular mount, and then it hung, so they tried again, causing another server to be spun up below? Would one of the two eventually get resolved, but not the other? Should there be a time limit on how long we listen and serve below?

cli.go
```diff
@@ -0,0 +1,502 @@
...
+	case "darwin":
+		cmd = "open"
+	default: // "linux", "freebsd", "openbsd", "netbsd"
+		cmd = "xdg-open"
+	}
+	args = append(args, url)
+	return exec.Command(cmd, args...).Start()
+}
+
+// parseError converts error from the API into summary and detailed portions.
```
tyrannosaurus-becks: Would it be possible here to leave in the comment an example of the error before and after?

cli.go
```diff
@@ -0,0 +1,502 @@
...
+		if detail == "" {
+			detail = errorParts[1]
+		}
+	}
+
+	return summary, detail
+
+}
+
+func errorHTML(summary, detail string) string {
```
tyrannosaurus-becks: Since these are so long and not Go, would it be possible to drop them into separate files? It would be super intuitive to name them something like `error.html.tmpl` and to consider using [Go's native template parsing](https://golang.org/pkg/text/template/).

path_oidc.go
```diff
@@ -17,6 +17,12 @@ import (
 
 var oidcStateTimeout = 10 * time.Minute
 
+// OIDC error prefixes. This are searched for specifically by the UI, so any
+// changes to them must be aligned with a UI change.
```
tyrannosaurus-becks: Nice! Thank you for leaving this note.

cli.go
kalafut: I've refactored things so the empty check now occurs in the `fetchAuthURL` method, returning an error if it fails. That should hopefully improve the consistency with the name.

cli.go
kalafut: The user can still copy/paste or click (if the terminal supports it) on the URL to complete the login. I think there are probably quite a few ways that an individual user's automatic browser opening could be borked, so we want them to still be able to complete the login.

cli.go
kalafut: It does! 

```
	// If WriteHeader is not called explicitly, the first call to Write
	// will trigger an implicit WriteHeader(http.StatusOK).
	// Thus explicit calls to WriteHeader are mainly used to
	// send error codes.
```

cli.go
kalafut: Renamed to `sigintCh`

cli.go
kalafut: It's a good question. What would happen currently is not good: the port would be in use which would generate an error but just hang there. I've updated it so now the listener must successfully start before the browser is opened, and any server error will be sent to the done channel to shut things down.

cli.go
kalafut: Yep

cli.go
kalafut: Bare template files don't work too well here since they'd have to be loaded at runtime but aren't bundled with the binary. So you're back to Go files for the strings. But I do agree that they're not like the rest of the CLI Go code and have moved them into their own file.

kalafut:
@tyrannosaurus-becks Addressed (hopefully!) all of your comments.

cli.go
tyrannosaurus-becks: :boom:

cli.go
tyrannosaurus-becks: Sounds good! Thank you!

cli.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	}
+
+	return summary, detail
+}
+
+// Help method for OIDC cli
+func (h *CLIHandler) Help() string {
+	help := `
+Usage: vault login -method=oidc [CONFIG K=V...]
```
tyrannosaurus-becks: Hey! This code is looking really good - the tests pass and the binary builds.

I've been working on manually giving this a spin locally and I'm not quite doing it right yet. Here's what I'm doing so far.
```
$ vault write sys/plugins/catalog/auth/custom-jwt \
    sha256=e0c59922a6a3db23f21e2ff848a372c09bbebe54959857c1ec46f2deb34918cd \
    command="vault-plugin-auth-jwt"
Success! Data written to: sys/plugins/catalog/auth/custom-jwt

$ vault auth enable custom-jwt
Success! Enabled custom-jwt auth method at: custom-jwt/

$ vault login -method=oidc role=engineering
Unknown auth method: oidc. Use "vault auth list" to see the complete list of
auth methods. Additionally, some auth methods are only available via the HTTP
API.

$ vault login -method='custom-jwt' role=engineering
Unknown auth method: custom-jwt. Use "vault auth list" to see the complete
list of auth methods. Additionally, some auth methods are only available via
the HTTP API.
```
Is that happening because I haven't added a role yet?

kalafut: Are you running Vault from the 1.1 beta branch? The recent change there adds the oidc auth method (for which the implementation lives here, making testing change a bit of a nuisance).

tyrannosaurus-becks: Oh I'm not! That must be it, I was on master.

tyrannosaurus-becks: I still get funny results, though different. After enabling the `custom-jwt` engine as shown in my previous comment, here's what I get.
```
$ vault login -method=oidc role=engineering
Error authenticating: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/oidc/oidc/auth_url
Code: 404. Errors:

* no handler for route 'auth/oidc/oidc/auth_url'

$ vault login -method=jwt role=engineering
Unknown auth method: jwt. Use "vault auth list" to see the complete list of
auth methods. Additionally, some auth methods are only available via the HTTP
API.

$ vault login -method=custom-jwt role=engineering
Unknown auth method: custom-jwt. Use "vault auth list" to see the complete
list of auth methods. Additionally, some auth methods are only available via
the HTTP API.

$ vault auth list
Path           Type          Accessor                    Description
----           ----          --------                    -----------
custom-jwt/    custom-jwt    auth_custom-jwt_b0fb6e78    n/a
token/         token         auth_token_a318d668         token based credentials
```
Do the instructions here need to be updated in some way or am I just going about things incorrectly?

kalafut: https://github.com/hashicorp/vault/pull/6222

kalafut: Lots of docs to write... In this case since you're not using the default "oidc" path, you'd need to add `-path=custom-jwt`

PR #24, opened 2019-02-13
Add OIDC support

### LINKED ISSUE #6: Login without providing role
For now one of the parameter to login endpoint id "role". For our use case, it will be better to drop this parameter. Without this parameter, the plugin must decide what policy to bind to the new token.
To do this properly the role must have a place to fill some binding to user/groups.

@jefferai Is it sane enhancements, WDYT?

kalafut:
This PR adds OIDC support to the plugin as well as a CLI login helper that can be imported by Vault. It also adds for both OIDC/JWT:

- default role support (fixes #6)
- the ability to copy claims into metadata, whether using OIDC or the existing JWT flow. (see discussion on https://github.com/hashicorp/vault-plugin-auth-jwt/pull/14)

It removes the "group_delimiter" method of specifying claims, replacing it with  [JSONPointer](https://tools.ietf.org/html/rfc6901).


claims.go
```diff
@@ -0,0 +1,65 @@
...
+			logger.Warn(fmt.Sprintf("unable to locate %s in claims: %s", claim, err.Error()))
+			return nil
+		}
+	}
+
+	// The claims unmarshalled by go-oidc don't use UseNumber, so there will
+	// be mismatches if they're coming in as float64 since Vault's config will
+	// be represented as json.Number. If the operator can coerce claims data to
+	// be in string form, there is no problem. Alternatively, we could try to
+	// intelligently convert float64 to json.Number, e.g.:
```
jefferai: go-oidc is actually quite small, it may be worth simply forking it so that we don't have to hack around this. If they change it upstream it changes the behavior so my guess is they won't want to do that, but we could ask. If we do fork we could put in a note essentially linking to the upstream and saying please don't rely on this library.

cli.go
```diff
@@ -0,0 +1,212 @@
...
+		return s.secret, s.err
+	case <-sigintCh:
+		return nil, errors.New("interrupted")
+	}
+}
+
+func fetchAuthURL(c *api.Client, role, mount, port string) (string, error) {
+	data := map[string]interface{}{
+		"role":         role,
+		"redirect_uri": fmt.Sprintf("http://localhost:%s/ui/vault/auth/%s/oidc/callback", port, mount),
```
jefferai: Is there a reason for this path to be so long? My worry is that it won't actually match the path used by a user (it doesn't, for instance, take into account namespaces) so will be confusing. It could just be `localhost:%s/oidc/callback` right?

Speaking of which -- I assume this will support namespaces via simply using the API client but may be good to check.

cli.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// becomes:
+//
+//    "No response from provider.", "Gateway timeout from upstream proxy."
+func parseError(err error) (string, string) {
+	headers := []string{errNoResponse, errLoginFailed, errTokenVerification}
+	summary := "Login error"
+	detail := ""
+
+	re := regexp.MustCompile(`(?s)Errors:.*\* *(.*)`)
```
jefferai: Usually must compiles are put into init() functions so that you don't get somewhere along the path of a request and *then* panic. I'd suggest using that pattern here.

cli.go
```diff
@@ -0,0 +1,212 @@
...
+			doneCh <- loginResp{nil, err}
+		}
+	}()
+
+	// Wait for either the callback to finish or SIGINT to be received
+	select {
+	case s := <-doneCh:
+		return s.secret, s.err
+	case <-sigintCh:
+		return nil, errors.New("interrupted")
```
jefferai: Since this is a CLI error technically this should probably be capitalized

cli.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	port, ok := m["port"]
+	if !ok {
+		port = defaultPort
+	}
+
+	role := m["role"]
+	if role == "" {
+		return nil, errors.New("a 'role' must be specified")
```
jefferai: This will error out even if default_role is specified in the API.

kalafut: Yep. I'll remove the `if`. The API will handle any rejection.

cli.go
kalafut: Fixed.

cli.go
kalafut: For the purposes of the CLI listener, it could anything. It's currently set up to match the pattern used by the UI so an operator can set up two redirect_uris that are the same except for the host portion. e.g. to provide both UI and CLI login ability, both of these would need to be in both the provider config and the Vault config.

https://vault-server.bigcorp.com:8200/ui/vault/auth/jwt/oidc/callback
http://localhost:8300/ui/vault/auth/jwt/oidc/callback

Does the similarity help, hurt? I can go either way on this one. Looking at them together, something really simple/short for localhost is probably better.

Yes I was relying on the API client to make NS work but it verify it is.

cli.go
kalafut: Moved regex compilation outside of the request path.

path_role.go
```diff
@@ -236,6 +283,15 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		role = new(jwtRole)
 	}
 
+	roleType := data.Get("role_type").(string)
+	if roleType == "" {
+		roleType = "jwt"
```
jefferai: I think this should be the other way around. The vast majority of users are likely to want OIDC behavior (and I think we'll be aliasing this plugin as `oidc` within Vault) so if not specified my suggestion is to default to OIDC.

path_oidc.go
```diff
@@ -0,0 +1,303 @@
...
+	uuid "github.com/hashicorp/go-uuid"
+	"github.com/hashicorp/vault/helper/strutil"
+	"github.com/hashicorp/vault/logical"
+	"github.com/hashicorp/vault/logical/framework"
+	"golang.org/x/oauth2"
+)
+
+var oidcStateTimeout = 10 * time.Minute
+
+// OIDC error prefixes. This are searched for specifically by the UI, so any
```
jefferai: pedantry: s/This/These

Gopkg.toml
```diff
@@ -61,6 +61,11 @@
   name = "gopkg.in/square/go-jose.v2"
   version = "2.1.8"
 
+# TODO: Remove this once https://github.com/hashicorp/go-plugin/pull/97 is merged
```
briankassouf: This has been merged now

path_config.go
```diff
@@ -103,6 +124,9 @@ func (b *jwtAuthBackend) pathConfigRead(ctx context.Context, req *logical.Reques
 		Data: map[string]interface{}{
 			"oidc_discovery_url":     config.OIDCDiscoveryURL,
 			"oidc_discovery_ca_pem":  config.OIDCDiscoveryCAPEM,
+			"oidc_client_id":         config.OIDCClientID,
+			"oidc_client_secret":     config.OIDCClientSecret,
```
briankassouf: So in many places we've stopped returning the root secret on read, perhaps we should remove it here too?

path_login.go
```diff
@@ -277,6 +203,120 @@ func (b *jwtAuthBackend) pathLoginRenew(ctx context.Context, req *logical.Reques
 	return resp, nil
 }
 
+func (b *jwtAuthBackend) verifyToken(ctx context.Context, config *jwtConfig, role *jwtRole, rawToken string) (map[string]interface{}, error) {
```
briankassouf: To distinguish this from the jwt signing verification method could we rename to verifyOIDCToken?

path_oidc.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	redirectURI := d.Get("redirect_uri").(string)
+	if redirectURI == "" {
+		return logical.ErrorResponse("missing redirect_uri"), nil
+	}
+
+	role, err := b.role(ctx, req.Storage, roleName)
+	if err != nil {
+		logger.Warn("error loading role", "error", err)
```
briankassouf: I don't think we need to log here or below

path_role.go
```diff
@@ -119,12 +160,15 @@ type jwtRole struct {
 	Period time.Duration `json:"period"`
 
 	// Role binding properties
-	BoundAudiences              []string                      `json:"bound_audiences"`
-	BoundSubject                string                        `json:"bound_subject"`
-	BoundCIDRs                  []*sockaddr.SockAddrMarshaler `json:"bound_cidrs"`
-	UserClaim                   string                        `json:"user_claim"`
-	GroupsClaim                 string                        `json:"groups_claim"`
-	GroupsClaimDelimiterPattern string                        `json:"groups_claim_delimiter_pattern"`
```
briankassouf: What are the repercussions of removing this?

path_role.go
briankassouf: agreed

path_oidc.go
kalafut: Removed

path_oidc.go
kalafut: Fixed

path_role.go
kalafut: The same capability can be achieved by revising the GroupsClaim to use a JSONPointer. This is definitely a backwards incompatibility, but I discussed with Jeff and he felt the user base of this fairly new field would be tiny, and the conversion to a JSONPointer is straightforward. This will be prominently called out in release notes/docs.

The immediate impact of doing nothing if you're using `GroupsClaimDelimiterPattern` is that group assignments won't happen.

path_role.go
kalafut: I agree re: the user preference, though the reason for the current approach is backwards compatibility. Current users aren't specifying a role type during configuration so they'll need to add `role_type=jwt` to any existing configuration scripts. If that is OK as an upgrade requirement, I have no issues changing it.

path_config.go
kalafut: Sure.

path_login.go
kalafut: Renamed

Gopkg.toml
kalafut: Constraint removed.

path_role.go
kalafut: Thinking about this more, I'm fine with the proposed change.

path_role.go
briankassouf: Sounds fine to me

cli.go
kalafut: Replaced with `localhost:%s/oidc/callback`

claims.go
kalafut: I've created a task to track resolution of this. Since the handling works with strings, I don't think it need block the 1.1 beta. Handling numbers will be an improvement.

PR #34, opened 2019-03-29
Enforce bound_cidrs in OIDC flow

kalafut:


path_oidc_test.go
```diff
@@ -361,8 +373,9 @@ func TestOIDC_Callback(t *testing.T) {
 			},
 		}
 		auth := resp.Auth
-		if diff := deep.Equal(auth, expected); diff != nil {
-			t.Fatal(diff)
+
+		if !reflect.DeepEqual(auth, expected) {
```
ncabatoff: You find this more readable?

kalafut: When comparing objects with *sockaddr.SockAddrMarshaler, deep will always return a diff.  I dug into this quite a while ago and don’t remember the exact details, but it has to do with comparing pointers versus pointed-to values, and how we are embeddding SockAddr. I just know DeepEqual does the comparison correctly and started using it in these rare cases.

kalafut:
I’ll add a test for the unpopulated case.

path_oidc_test.go
ncabatoff: Ah, that makes sense.  I'll have to try go-cmp at some point and see if it does any better.

PR #37, opened 2019-04-01
Escape URL in CLI helper on Windows

### LINKED ISSUE #36: oidc redirect is not successfull on Windows
On Linux/OSX when I use "vault login -method=oidc role=role, I'm redirected to the IDP and the login is successfull. 
On Windows (10) the url is truncated (whatever the browser that is called) before the first ampersand.

Ex: 
`https://the.idp.com/adfs/oauth2/authorize/?client_id=xxxx&nonce=xxx&redirect_uri=http%3A%2F%2Flocalhost%3A8250%2Foidc%2Fcallback&response_type=code&scope=openid&state=xxx`
is truncated in the browser url field to : 
`https://the.idp.com/adfs/oauth2/authorize/?client_id=xxxx`
Thus the login attempt fail.

kalafut:
Fixes #36

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -133,6 +133,7 @@ func openURL(url string) error {
 	case "windows":
 		cmd = "cmd"
 		args = []string{"/c", "start"}
+		url = strings.Replace(url, "&", "^&", -1)
 	case "darwin":
 		cmd = "open"
 	default: // "linux", "freebsd", "openbsd", "netbsd"


kalafut:
`strings.ReplaceAll` is a Go 1.12 feature I'd like to use, but I think we're still holding off on 1.12 for the time being.

PR #38, opened 2019-04-04
Disable /login path for roles with role_type oidc

bks7:
Vault must not accept signed JWT tokens through /login path when role
has role_type oidc, since there might be a situation when the client
secret could be compromised, and thus the malicious might be able to
illegitimately get a token with the right aud claim, which Vault would
accept through the /login path.


kalafut:
@bks7 Thanks for the PR. I think your suggestion makes sense. The current approach was in support of a Kubernetes use case that isn't needed at this point, so removing this ability is fine. I'll plan to get to the review soon.

kalafut:
Thanks!

PR #41, opened 2019-04-09
Interpret bound claims value as a "matches any of" test if a list is provided

kalafut:


claims.go
```diff
@@ -96,8 +96,27 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
...
-			return fmt.Errorf("claim %q does not match associated bound claim", claim)
+		var expVals []string
+
+		switch v := expValue.(type) {
+		case []string:
+			expVals = v
+		case string:
+			expVals = []string{v}
+		default:
+			return fmt.Errorf("bound claim not a string or []string: %v", expValue)
```
briankassouf:
```suggestion
			return fmt.Errorf("bound claim not a string or list: %v", expValue)
```

kalafut:
Discovered that this PR needs some updates to work in the real world (i.e. not just unit test context) due to how the `bound_claims` structure is being stored in the role.

PR #43, opened 2019-04-24
Add JWKS support

### LINKED ISSUE #27: Dynamically update verification key with plain JWT
The specification differs between JWT and OIDC IDToken. For example, the `iss` claim is **not** mandatory in JWT, but it is mandatory in IDToken.
Currently, we need to use `oidc_discovery_url` to dynamically rotate the public key for signature verification. In that case, the JWT must satisfy the specification of ID Token.

Could you support the only jwks endpoint so that we can dynamically update the key in JWT(not OIDC) authentication?

kalafut:
Closes #27

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -6,9 +6,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/hashicorp/errwrap"
-
 	"github.com/coreos/go-oidc"
+	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/logical"
 	"github.com/patrickmn/go-cache"
--- a/backend.go
+++ b/backend.go
@@ -2,13 +2,16 @@ package jwtauth
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
-	oidc "github.com/coreos/go-oidc"
+	"github.com/hashicorp/errwrap"
+
+	"github.com/coreos/go-oidc"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/logical"
-	cache "github.com/patrickmn/go-cache"
+	"github.com/patrickmn/go-cache"
 )
 
 const (
@@ -30,6 +33,7 @@ type jwtAuthBackend struct {
 
 	l            sync.RWMutex
 	provider     *oidc.Provider
+	keySet       oidc.KeySet
 	cachedConfig *jwtConfig
 	oidcStates   *cache.Cache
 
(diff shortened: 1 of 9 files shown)


backend.go
```diff
@@ -2,13 +2,16 @@ package jwtauth
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
-	oidc "github.com/coreos/go-oidc"
+	"github.com/hashicorp/errwrap"
```
tyrannosaurus-becks: Probably should be moved down to go with the other hashicorp imports.

path_config.go
```diff
@@ -160,12 +180,32 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
...
+
+		// Try to verify a correctly formatted JWT. The signature will fail to match, but other
+		// errors with fetching the remote keyset should be reported.
+		testJWT := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.Hf3E3iCHzqC5QIQ0nCqS1kw78IiQTRVzsLTuKoDIpdk"
+		_, err = keyset.VerifySignature(ctx, testJWT)
+		if err == nil {
+			err = errors.New("unexpected verification of JWT")
+		}
+
+		if !strings.Contains(err.Error(), "failed to verify id token signature") {
```
tyrannosaurus-becks: Question, where does this error string come from? Is it coming from our code or some remote API of the user's somewhere?

path_oidc.go
```diff
@@ -215,13 +215,14 @@ func (b *jwtAuthBackend) authURL(ctx context.Context, req *logical.Request, d *f
...
 	if err != nil {
-		logger.Warn("error loading configuration", "error", err)
-		return resp, nil
+		return nil, err
 	}
-
 	if config == nil {
-		logger.Warn("nil configuration")
-		return resp, nil
+		return logical.ErrorResponse("could not load configuration"), nil
```
tyrannosaurus-becks: Just for my own education, what is the value of returning `nil, err` in some places, versus `errResp, nil` in other places? I actually don't know and I've tended toward the first.

path_config.go
kalafut: It is from the library we use. I don't love this approach, but it's not too bad given that it's not an API response. If they change this message, our tests will fail which is 👍 .

backend.go
kalafut: Yep

path_oidc.go
kalafut: Broadly (as I know it), it has to do with 400s (err,nil) vs 500s (nil, err). In this case, a config being nil is something the user controls pretty directly, but the previous case of a storage error is more of a 500/"Internal Service" type error. But that's just observation of how the pattern has been used, and there is variation. But it definitely does effect whether the resp code is a 4xx or 5xx.

kalafut:
Running concurrency tests is a good idea and something I'm going to put in.

path_oidc.go
tyrannosaurus-becks: Ah! That is super helpful! Thanks!

kalafut:
All comments should be addressed.

path_login.go
```diff
@@ -81,29 +82,51 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+			if err != nil {
+				return logical.ErrorResponse(errwrap.Wrapf("error verifying token: {{err}}", err).Error()), nil
+			}
+
+			// Unmarshal payload into two copies: public claims for library verification, and a set
+			// of all received claims.
+			if err := json.Unmarshal(payload, &claims); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal claims: %v", err)
+			}
+			if err := json.Unmarshal(payload, &allClaims); err != nil {
```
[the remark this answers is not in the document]
kalafut: `claims` and `allClaims` are different types. Copying would need to be explicitly for each member, which is why I just left the second unmarshal.

PR #45, opened 2019-05-07
Update pointerstructure module

tam7t:
github.com/mitchellh/pointerstructure was updated to include a license. This PR just updates the module and re-vendors the package now complete with the LICENSE file.


kalafut:
Thanks!

PR #49, opened 2019-05-09
Improve role config jwt bound constraint check

shwuandwing:
When configuring JWT roles, consider bound claims when checking if there is are bound constraints


shwuandwing:
#49

shwuandwing:
@kalafut , this is my first Github change.  Is there any more steps I have to take to get this change approved?

kalafut:
@shwuandwing Hi. When I last looked I didn't see the CLA approval, but now that's in good shape so I'll be doing the review soon.

kalafut:
Good catch and thanks for the submission!

PR #50, opened 2019-05-10
Add functionality to check against a list of claims

MartinLeedotOrg:
If a list of claims is provided, unmarshal it and then search within that for the bound claim.

The intention is to allow me to pass a claim like:
```vault_role: ["admin", "superadmin"]```

This will allow me to log in as either `admin` or `superadmin` as I choose.

Useful for testing roles that other users have, or to allow people who sit across two teams to perform functions in both without adding additional roles.

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -1,6 +1,7 @@
 package jwtauth
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -107,11 +108,22 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
 			return fmt.Errorf("bound claim is not a string or list: %v", expValue)
 		}
 
+		var actVals []string
+		actData := []byte(actValue.(string))
+
+		err := json.Unmarshal(actData, &actVals)
+		if err != nil {
+			actVals = []string{actValue.(string)}
+		}
+
+
 		found := false
 		for _, v := range expVals {
-			if actValue == v {
-				found = true
-				break
+			for _, av := range actVals {
+				if av == v {
+					found = true
+					break
+				}
 			}
 		}
 
--- a/claims.go
+++ b/claims.go
@@ -1,7 +1,6 @@
 package jwtauth
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -97,7 +96,16 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
 			return fmt.Errorf("claim %q is missing", claim)
 		}
 
-		var expVals []interface{}
+		var actVals, expVals []interface{}
+
+		switch v := actValue.(type) {
+		case []interface{}:
+			actVals = v
+		case string:
+			actVals = []interface{}{v}
+		default:
+			return fmt.Errorf("received claim is not a string or list: %v", actValue)
+		}
 
 		switch v := expValue.(type) {
 		case []interface{}:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by chrissphinx** [comment]
@MartinLeedotOrg have you confirmed this is working? I tried to build `vault` from source with this patch applied, but I am still unable to authenticate despite being in the GitLab group I am checking for.

I tried both:
```
  "bound_claims": {
    "groups": "group1"
  }
```

and:
```
  "bound_claims": {
    "groups": [ "group1" ]
  }
```

without any success. 😞

**Comment by MartinLeedotOrg** [comment]
Did the first case work before the patch?

I haven't tested this with Gitlab's groups - only with Okta and passing arbitrary claims.

Are you requesting the groups scope from Gitlab?

**Comment by kalafut** [comment]
@MartinLeedotOrg Thanks! The JSON parsing is all handled upstream so I've adjust that bit and added a couple more tests. I just tested this against Gitlab groups and it now appears to work OK.

@chrissphinx Maybe give this update a try?

File: claims.go [hunk 1]
```diff
@@ -108,10 +117,14 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
...
 
 		found := false
+
+	scan:
 		for _, v := range expVals {
-			if actValue == v {
-				found = true
-				break
+			for _, av := range actVals {
+				if av == v {
```
tyrannosaurus-becks: Will this match if one or both are an interface with the same underlying value? Not sure if that's what's happening here.

tyrannosaurus-becks: Actually, I see the tests prove that it does match when it's supposed to. I'll resolve this!

chrissphinx:
@kalafut no, unfortunately still not working for me. Perhaps I have built the binary wrong, I suppose we will wait for this to come through on official release to test out then.

kalafut:
@chrissphinx Can you post your gitlab role config here (w/o sensitive data)?

chrissphinx:
Sure it's:

```
{
  "user_claim": "sub",
  "policies": "employee",
  "bound_claims": {
    "groups": "group1"
  },
  "allowed_redirect_uris": [ "https://<vault_url>/ui/vault/auth/oidc/oidc/callback" ]
}
```

I have also tried it like:

```
{
  "user_claim": "sub",
  "policies": "employee",
  "bound_claims": {
    "groups": [ "group1" ]
  },
  "allowed_redirect_uris": [ "https://<vault_url>/ui/vault/auth/oidc/oidc/callback" ]
}
```

kalafut:
@chrissphinx The plugin hasn't been updated in the Vault repo yet which is why you're not seeing the change. To test it out now you'd need to build the plugin binary and register it with Vault. We'll merge the plugin into the main binary when we approach the next release.

No, you should not have to remake the OIDC auth method you've set up. You can always choose to mount another one at separate path if you want to run tests first, however.

chrissphinx:
@kalafut yeah so what I did was clone down the main `vault` repo and then remove the `vault-plugin-auth-jwt` folder and clone down _this_ repo in it's place. Then I built the binary. Maybe that process will not work?

Looking forward to the next release then, thanks for all the hard work! 👍

kalafut:
@chrissphinx That process would work if you're not using Go modules, so it depends. If Go modules are used, then then `vendor` folder is ignored and the Go module cache is used. To do the same hack but for Go modules you could modify `go.mod` to point to a specific commit for jwt repo (since it hasn't been tagged yet. details: https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies).  Then go build will fetch that version.

PR #53, opened 2019-06-13
Add configurable expiration/notbefore leeways

jasonodonnell:
Leeways for expiration and not before were hardcoded to 300s (5m) for logins.  These are now configurable (with a default of 150s) via two new args: `expiration_leeway` and `not_before_leeway`.

Additionally added `clock_skew_leeway` to configure a 60s (1m) leeway time for _all_ claims.  This can be toggled off if no extra leeway is required.


path_role.go
```diff
@@ -70,6 +70,18 @@ should never expire. The token should be renewed within the
 duration specified by this value. At each renewal, the token's
 TTL will be set to the value of this parameter.`,
 			},
+			"expiration_leeway": {
+				Type: framework.TypeDurationSecond,
+				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
+Defaults to 300 (5 minutes).`,
+				Default: "300",
```
kalafut: Probably want to avoid defaulting here and for `not_before_leeway` so we don't persist this value (which in turn let's us change it for those who haven't explicitly set it).

jasonodonnell: I think this is a good idea but it makes it a little awkward because the role configuration says one thing (IE leeways are 0) but logins do another thing (default to 300 seconds).  Do you think this might be confusing to users?

jasonodonnell: @kalafut Added your suggestion for review.

path_login_test.go
```diff
@@ -666,13 +670,59 @@ func testLogin_JWT(t *testing.T, jwks bool) {
...
+	}
+
+	// test custom expiration/notbefore from issue time
+	{
+		b, storage := setupBackend(t, false, false, true, false, false, jwks, true)
+
+		cl := jwt.Claims{
+			Subject:  "r3qXcK2bix9eFECzsU3Sbmh0K16fatW6@clients",
+			Issuer:   "https://team-vault.auth0.com/",
+			IssuedAt: jwt.NewNumericDate(time.Now()),
```
kalafut: To cover this change I think you'd want a couple more test cases. This one is a good/valid case (though a time the was near--but within--the limit would be better), but I don't see bad/invalid cases for when the current time is before the calculated nbf or after the calculated exp.

path_login.go
```diff
@@ -176,7 +177,9 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 			Time:    time.Now(),
 		}
 
-		if err := claims.Validate(expected); err != nil {
+		// We bake leeway into expiration/not before, so we pass in a duration of 0
```
kalafut: Good catch about the double-leeway, but that will only apply if our calculations are used. If the JWT has a `iat`/`nbf` and `exp`, we'll now be checking it with no leeway, when before the standard library amount was used. We should probably have a default value set to `jwt.DefaultLeeway`, and then reset that to 0 if we end up doing out our exp/nbf calcs.

path_login.go
```diff
@@ -167,7 +178,15 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+		// to account for our implementation.  Default to JWT's package of 1 minute
+		// if no leeways are configured
+		leeway := time.Second * 0
```
jefferai: A more canonical way to do this would be `var leeway time.Duration`

path_login.go
```diff
@@ -167,7 +178,15 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+		if role.ExpirationLeeway == 0 && role.NotBeforeLeeway == 0 {
+			leeway = jwt.DefaultLeeway
```
jefferai: This feels odd to me -- this means that if you don't set either of those, we're adding in 5 minutes of leeway above (in each direction), but then another minute here, so it's actually a 12 minute validity period. Granted this leeway accounts for clock skew in either direction, which the above calculations don't, but in the same vein, if you define the leeways above, or if you have `iat`/`nbf` defined, now you're *not* accounting for clock skew.

I wonder if the right logic is that we should always use the library's default leeway value to account for clock skew, but reduce the defaults we use above from 300 seconds down to, say, 150 seconds for each. That gives a validity window of 300 seconds total plus leeway.

If we add leeway all the time we should document it in CHANGES as it's a behavioral change. It may be worth another parameter, to be honest -- something like `clock_skew_leeway` that gets applied regardless of claims and in both directions, where we default to the library's default but people can turn it off if they don't want an extra minute of validity (which they may well not want).

path_role.go
```diff
@@ -69,6 +69,21 @@ be renewed. Defaults to 0, in which case the value will fall back to the system/
...
+				Type: framework.TypeDurationSecond,
+				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
+Defaults to 150 (2.5 minutes).`,
+			},
+			"not_before_leeway": {
+				Type: framework.TypeDurationSecond,
+				Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew. 
+Defaults to 150 (2.5 minutes).`,
+			},
+			"clock_skew_leeway": {
```
jefferai: Why a bool instead of a duration? Based on past issues and feedback different users have different clock skew tolerances.

jasonodonnell: I can make this a duration if you want but I thought we just wanted to toggle the JWT package default on and off based on your last review:

> ... where we default to the library's default but people can turn it _off if they don't want an extra minute of validity (which they may well not want)._

jefferai: Thanks for changing this. I had simply meant using its value as the default. Note that you can(/should) put a `Default` for each of these, which then allows that to show up in OpenAPI. For the clock skew leeway the value could just be `jwt.DefaultLeeway` or whatever it's called.

path_role.go
```diff
@@ -70,6 +73,24 @@ should never expire. The token should be renewed within the
 duration specified by this value. At each renewal, the token's
 TTL will be set to the value of this parameter.`,
 			},
+			"expiration_leeway": {
+				Type: framework.TypeDurationSecond,
+				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
+Defaults to 150 (2.5 minutes), minimum of 1 second.`,
```
jefferai: Post beta1, can you add text here (and in the other options you added) indicating that setting to 0 will use the default value?

path_role.go
```diff
@@ -70,6 +73,24 @@ should never expire. The token should be renewed within the
...
+				Default: claimDefaultLeeway,
+			},
+			"not_before_leeway": {
+				Type: framework.TypeDurationSecond,
+				Description: `Duration in seconds of leeway when validating not before values of a token to account for clock skew. 
+Defaults to 150 (2.5 minutes), minimum of 1 second..`,
```
jefferai: dub period

PR #54, opened 2019-06-18
Restore client_id check

kalafut:
This check was incorrectly removed for OIDC roles in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/38.

The PR also includes a test for this case and some general test cleanup.


kalafut:
> This looks good, though of course the compilation error will need to be solved first. Looks like possibly a dependency issue?

travis hiccup, I guess. I just press rebuild and it was fine.

PR #57, opened 2019-06-26
Add option to allow verbose OIDC logging

kalafut:


path_oidc.go
```diff
@@ -161,6 +165,12 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 		logFunc("error reading /userinfo endpoint", "error", err)
 	}
 
+	if role.VerboseOIDCLogging {
+		if c, err := json.Marshal(allClaims); err == nil {
```
tyrannosaurus-becks: Would this ever fail to marshal? If it could, would it be useful to log the raw claims because of their unexpected state? Part of me thinks this would be super unlikely to fail, and would just return empty bytes anyways, so maybe in this situation it would be ok to just drop the err and log the potentially empty bytes.

path_role.go
```diff
@@ -128,6 +129,9 @@ authenticate against this role`,
 				Type:        framework.TypeCommaStringSlice,
 				Description: `Comma-separated list of allowed values for redirect_uri`,
 			},
+			"verbose_oidc_logging": {
+				Type: framework.TypeBool,
```
tyrannosaurus-becks: Since this will be perpetuated to the UI, might be nice here to add a description that includes a warning that it causes the logging of sensitive information. Not the kind of stuff you'd want picked up by filebeat and sent off to ELK, probably.

path_oidc.go
kalafut: I'd put this at a near 0% chance, since `allClaims` is the result of json.Unmarshalling just a few lines earlier. But I will add a change to just dump the error to the log if by change there is one.

path_role.go
kalafut: Yeah. I'm going to add a note here, and probably a warning in the config write response if you're enabling this.

PR #58, opened 2019-06-26
Change leeways to TypeSignedDurationSecond

jasonodonnell:
Converting the custom leeway parameters to use the new `TypeSignedDurationSecond` type to support disabling leeway.

### DIFF
--- a/path_role.go
+++ b/path_role.go
@@ -74,21 +74,21 @@ duration specified by this value. At each renewal, the token's
 TTL will be set to the value of this parameter.`,
 			},
 			"expiration_leeway": {
-				Type: framework.TypeDurationSecond,
+				Type: framework.TypeSignedDurationSecond,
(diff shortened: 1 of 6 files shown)


path_role.go
```diff
@@ -74,21 +74,21 @@ duration specified by this value. At each renewal, the token's
 TTL will be set to the value of this parameter.`,
 			},
 			"expiration_leeway": {
-				Type: framework.TypeDurationSecond,
+				Type: framework.TypeSignedDurationSecond,
 				Description: `Duration in seconds of leeway when validating expiration of a token to account for clock skew. 
-Defaults to 150 (2.5 minutes), minimum of 1 second.`,
+Defaults to 150 (2.5 minutes) if set to 0 and can be disabled if set to -1.`,
```
mgaffney: Is there any value in adding tests for negative values to `path_role_test.go`?

jasonodonnell: Definitely valuable to have so I added it 👍

PR #59, opened 2019-07-02
Tokenutilize

jefferai:


path_role_test.go
```diff
@@ -613,6 +637,10 @@ func TestPath_Read(t *testing.T) {
 			t.Fatal("unexpected bound cidrs")
 		}
 		delete(resp.Data, "bound_cidrs")
+		if resp.Data["token_bound_cidrs"].([]*sockaddr.SockAddrMarshaler)[0].String() != "127.0.0.1/8" {
```
jasonodonnell: Should we check if `len(resp.Data["token_bound_cidrs"]) > 0` before running this test to avoid out of bounds errors?

jefferai: Add if you like. It's a test, so a panic in the test is a fine way to know something is wrong rather than an explicit t.Fatal :-)

PR #61, opened 2019-07-19
Forward OIDC logins received on performance standbys

kalafut:
Addresses `expired/missing oauth state` errors

### DIFF
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -8,15 +8,13 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


path_oidc.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/vault/sdk/helper/consts"
+
```
calvn: Extra newline here

kalafut: Fixed imports

PR #63, opened 2019-07-26
Convert a single string groups claim into a list of one

### LINKED ISSUE #62: ADFS 4.0 groups claim could not be converted to list
We have OIDC configured to work with ADFS 4.0 and we are hitting an issue when we want to use `groups_claim`. What happens is that ADFS issues the groups claim as `string` if there is only one group that the user belongs to. If there are multiple groups, then it sends the groups claim as an `array` which Vault accepts.

We are hitting the following error: https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_login.go#L345 because Vault expects the `groups` claim to always be an `array`.

I tried to force ADFS to send the groups claim as an array even if there is only one group, but I couldn't figure out how to do it.

Is this something that can be fixed on Vault's side?

kalafut:
This also refactors a crazy long test function signature into
a config structure.

Fixes #62

### DIFF
--- a/claims_test.go
+++ b/claims_test.go
@@ -423,10 +423,10 @@ func Test_normalizeList(t *testing.T) {
 	for _, tt := range tests {
 		normalized, ok := normalizeList(tt.raw)
 		if !reflect.DeepEqual(normalized, tt.normalized) {
-			t.Errorf("normalizeList() got = %v, want %v", normalized, tt.normalized)
+			t.Errorf("normalizeList() got normalized = %v, want %v", normalized, tt.normalized)
 		}
 		if ok != tt.ok {
-			t.Errorf("normalizeList() got1 = %v, want %v", ok, tt.ok)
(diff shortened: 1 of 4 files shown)


claims_test.go
```diff
@@ -376,3 +377,56 @@ func TestValidateBoundClaims(t *testing.T) {
...
+			ok:         false,
+		},
+	}
+	for _, tt := range tests {
+		normalized, ok := normalizeList(tt.raw)
+		if !reflect.DeepEqual(normalized, tt.normalized) {
+			t.Errorf("normalizeList() got = %v, want %v", normalized, tt.normalized)
+		}
+		if ok != tt.ok {
+			t.Errorf("normalizeList() got1 = %v, want %v", ok, tt.ok)
```
jasonodonnell: Is this `got1` a typo or are you trying to have a different error message for this case?

kalafut: oops. I used the GoLand test case generator which is convenient but has some stuff that I usually change, or in this case forget to change.

PR #78, opened 2019-10-29
Azure AD, more than 200 groups

### LINKED ISSUE #74: AzureAD/OIDC to support more than AD 200 groups
**Is your feature request related to a problem? Please describe.**
There is a limitation within AAD and OIDC when a user is in more then 200 groups within Azure AD the groups claim will not be sent instead the below is sent
```
"_claim_names": {
    "groups": "src1"
  },
  "_claim_sources": {
    "src1": {
      "endpoint": "https://graph.windows.net/{TENANT_ID}/users/{USER-GUID}/getMemberObjects"
    }
  }
```
https://azure.microsoft.com/en-us/resources/samples/active-directory-dotnet-webapp-groupclaims/ see under Groups overage claim

**Describe the solution you'd like**
To have the _claim_sources supported by vault

Gaardsholt:
This should fix #74

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -5,13 +5,17 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/coreos/go-oidc"
 	"github.com/hashicorp/errwrap"
+	log "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/helper/cidrutil"
 	"github.com/hashicorp/vault/sdk/logical"
+	"github.com/mitchellh/pointerstructure"
 	"gopkg.in/square/go-jose.v2/jwt"
 )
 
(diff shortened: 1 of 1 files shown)


Gaardsholt:
I signed the CLA yesterday, but it still says that I haven't done so?

wrighbr:
+1

Gaardsholt:
@kalafut I accidentally used my work email when I pushed those commits.
CLA has now be signed :)

Gaardsholt:
@kalafut Any news on this one? :)

kalafut:
@Gaardsholt Hoping to get to it soon. This one takes a little more time due to the environment setup. But some initial feedback after a quick glance: the PR looks fairly Azure-specific, but is it necessary to be that way? Ideally this would handle distributed claims from any provider, but I'm not familiar enough with Azure's implementation to know whether there are special cases due to them not following a spec or adding on their own steps. At which point this starts to veer into the issues I mentioned in #83.

Gaardsholt:
@kalafut After reading some of the documentation for GSuite OIDC setup, it seems like they are doing it slightly different.
So either we need to expand the check for all the different quirks that providers are doing.
Or as a part of the OIDC setup in Vault, have the option to specify which provider you are using.

tvoran:
Hi @Gaardsholt, we're continuing this one in #120 so I'll close this for now. It'd be great if you could take a look and see if it works for you.

Issue #79, opened 2019-10-31
Support for https and different port for callbackhost


gites:
For certain use cases, it would be nice to have the ability to use https in redirect_uri and set a different port for callback host than the port that Vault CLI uses for listening for the callback.

PR #80, opened 2019-10-31
more customisation for callback URL

gites:
Solves https://github.com/hashicorp/vault-plugin-auth-jwt/issues/79

* allow for setting `http` or `https` method in `redirect_uri`
* allow for setting port in `redirect_uri`
* updated plugin help info

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -20,7 +20,6 @@ const defaultMount = "oidc"
 const defaultPort = "8250"
 const defaultCallbackHost = "localhost"
 const defaultCallbackMethod = "http"
-const defaultCallbackPort = "8250"
 
 var errorRegex = regexp.MustCompile(`(?s)Errors:.*\* *(.*)`)
 
@@ -59,14 +58,14 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 		callbackMethod = defaultCallbackMethod
 	}
 
-	callbackport, ok := m["callbackport"]
+	callbackPort, ok := m["callbackport"]
 	if !ok {
-		callbackport = defaultCallbackPort
+		callbackPort = port
 	}
 
 	role := m["role"]
 
-	authURL, err := fetchAuthURL(c, role, mount, callbackport, callbackMethod, callbackHost)
+	authURL, err := fetchAuthURL(c, role, mount, callbackPort, callbackMethod, callbackHost)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


kalafut:
@gites Thanks for the contribution. I'm taking a look at this now. I think the customization is reasonable, though I was curious what use cases you envision for the callbackport parameter.

gites:
Thanks for your response. We are running the Vault cluster in private VPC with OIDC logins enabled, but it is a bit annoying to copy-paste and curl the localhost callback URL. So the idea here is to pass the callback through some external, internet-facing,  domain and small proxy (that will validate the callback request) to the Vault CLI on a designated host. It would be nice to do this over TLS and port 443.

cli.go
```diff
@@ -52,9 +54,19 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
 	}
 
+	callbackMethod, ok := m["callbackmethod"]
+	if !ok {
+		callbackMethod = defaultCallbackMethod
+	}
+
+	callbackport, ok := m["callbackport"]
+	if !ok {
+		callbackport = defaultCallbackPort
```
kalafut: I think this also needs to take into account whether `port` was set in order to maintain the existing behavior.

For example, this command will fail with the PR as is: `vault login -method=oidc role=test port=8400`. In the released version the redirect_uri is set to 8400, but with the PR it's now getting the default of 8250.

gites: Added a fix that will set `callbackport` to the value of `port` if no explicit `callbackport` option is passed.

gites:
@kalafut I think, the backwards compatibility issue is fixed.

DrDaveD:
I have been trying to get callbackmethod=https to work, but haven't been able to find a way to get vault login to listen with https.  Am I correct from the discussion here that it is only expected to work via an https reverse proxy?

I wish it would just work through a full time https listener, the one for the ui on port 8200.  We want to use one vault for multiple users, and more than one won't be able to listen on port 8250 at the same time.  Better yet, I wish it could use the [Oauth 2.0 device flow](https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15).

tomwganem:
I'm also confused by the https callback method. Is there a way to supply a cert?

DrDaveD:
For reference, I requested the device flow in #103.

kalafut:
Hi. You're correct that the https option (and some of others on the CLI) is just to allow the user to customize the redirect for more complex setups such as operating behind a proxy. 

Having the CLI terminate the TLS connection would be difficult in most cases since the callback is coming from the browser, and thus requires the private key to a browser-trusted credential chain, which probably isn't on users' devices.

Issue #85, opened 2019-11-20
OIDC callback listener binds to all IP addresses on the system by default


ianferguson:
The OIDC callback listener does not currently specify an IP address or name to bind the listener to: https://github.com/hashicorp/vault-plugin-auth-jwt/blob/716d11f2bbd51c390b2e0f7e31ff128bfa68a815/cli.go#L97 

The [golang net package documentation](https://golang.org/pkg/net/#Listen) specifies that not providing a host or ip address results in the following behavior:

> if the host in the address parameter is empty or a literal unspecified IP address, Listen listens on all available unicast and anycast IP addresses of the local system.

This results in users on OS X seeing a pop up dialog box asking `Do you want the application “vault” to accept incoming network connections?` when they execute `vault login -method=oidc` because the listener is attempting to bind to non-localhost ip's.

By default it might be preferable for most use cases to bind only to `localhost` by default, and update the code linked above to read `listener, err := net.Listen("tcp", "localhost:"+port)`.

Changing the value to `"localhost:"+port` may cause problems for the use cases #80 was added to support, so likely a `listenaddress` parameter should be made available so that those use cases can override a default `localhost` value.

I've opened PR #86 with a possible fix for this

kalafut:
Fixed in #86

PR #86, opened 2019-11-20
Default OIDC callback listener to only listen on localhost

ianferguson:
This addresses the issue raised in #85 in a way that should allow flexibility for users of the options added in #80.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -17,6 +17,7 @@ import (
 )
 
 const defaultMount = "oidc"
+const defaultListenAddress = "localhost"
 const defaultPort = "8250"
 const defaultCallbackHost = "localhost"
 const defaultCallbackMethod = "http"
@@ -41,6 +42,11 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 	mount, ok := m["mount"]
 	if !ok {
 		mount = defaultMount
+	}
+
+	listenAddress, ok := m["listenaddress"]
+	if !ok {
+		listenAddress = defaultListenAddress
 	}
 
 	port, ok := m["port"]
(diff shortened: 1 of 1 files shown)


cli.go
```diff
@@ -239,14 +245,17 @@ Configuration:
...
-	  Optional localhost port to use for OIDC callback (default: 8250).
+    Optional localhost port to use for OIDC callback (default: 8250).
 
   callbackmethod=<string>
-	  Optional method to to use in OIDC redirect_uri (default: http).
+    Optional method to to use in OIDC redirect_uri (default: http).
 
   callbackhost=<string>
-	  Optional callback host adddress to use in OIDC redirect_uri (default: localhost).
+    Optional callback host address to use in OIDC redirect_uri (default: localhost).
```
ianferguson: the help string had a mix of tabs and spaces towards the end, which I updated to be more consistently spaces

kalafut:
Trying to think if there is any aspect of this connection that is still not configurable 🙂.  But for completeness I agree this probably makes sense.

ianferguson:
@kalafut thank you for the quick reply!

My first instinct was to just hardcode "localhost" into the address field of the `net.Listen`, but given the recent additions in #80, I wanted to make sure @gites' use case didn't have any issues with redirecting to some remote host. 

Let me know if there is anything else I can provide to help!

ianferguson:
Is there anything I can do to help get this change merged in?

kalafut:
@ianferguson I think it looks good. I want to test a couple things (today-ish) but I don't have any PR requests.

kalafut:
Thanks!

ianferguson:
@kalafut thanks for doing that! Should I update the go.mod file in the hashicorp/vault project (https://github.com/hashicorp/vault/blob/15d31863801867f3b03a30c565439089830f675a/go.mod#L76) so that this is picked up in the next Vault release, or will y'all take care of that?

kalafut:
@ianferguson We'll handle that. Thanks.

ianferguson:
Sounds good, thanks again for your help!

Issue #82, opened 2019-11-04
Match claim values using wildcards or regular expressions


randomswdev:
Currnetly the auth backend requires exact matches on the claim values. But sometimes it could be useful to allow matching just on portions of the claim, like for example all the departments whose name starts with "Engineering-". I can for sure enumerate all such departments that exist today, but I will have to modify the configuration of the backend every time we define a new department.

Is there any intereste and/or plan to implement such an enhancement in the future? We can work also in contributing a change request if the feature can be of interest.

randomswdev:
Submitted a pull request to implement this change: https://github.com/hashicorp/vault-plugin-auth-jwt/pull/89

randomswdev:
The pull request has been merged. So I think this one can be closed. @kalafut what do you think?

kalafut:
Ah yeah. Thanks.

Closed by #89.

PR #89, opened 2019-12-06
Add the ability to match bound claims using globs

randomswdev:
This merge request is intended to add the ability to match bound claims not only using an exact string matching, but also a mathcing based on regular expressions.
Added a new flag `bound_claims_are_regexps`; if it is set to true the vaule of each entry in the `bound_claims` map is expected to be a string (not an array of strings) that will be interpreted as a regular expresison. The syntax for the regular expressions is the golang syntax.

This would allow mathcing JWT whose claims, for example, contain randomly generated portions of text.

Solves issue https://github.com/hashicorp/vault-plugin-auth-jwt/issues/82

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -116,10 +116,7 @@ func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims,
 	scan:
 		for _, v := range expVals {
 			if useGlobs {
-				vs, ok := v.(string)
-				if !ok {
-					return fmt.Errorf("received claim is not a glob string: %v", v)
-				}
+				vs := v.(string)
 				for _, av := range actVals {
 					if avs, ok := av.(string); ok {
 						if glob.Glob(vs, avs) {
--- a/claims.go
+++ b/claims.go
@@ -91,10 +91,6 @@ func validateAudience(boundAudiences, audClaim []string, strict bool) error {
 // validateBoundClaims checks that all of the claim:value requirements in boundClaims are
 // met in allClaims.
 func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims, allClaims map[string]interface{}) error {
-	if boundClaimsType != boundClaimsTypeString && boundClaimsType != boundClaimsTypeGlob {
-		return fmt.Errorf("claim type %s is not valid", boundClaimsType)
-	}
-
 	useGlobs := boundClaimsType == boundClaimsTypeGlob
 
 	for claim, expValue := range boundClaims {
--- a/claims.go
+++ b/claims.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/vault/sdk/helper/strutil"
+	"github.com/ryanuber/go-glob"
 
 	log "github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/pointerstructure"
@@ -89,7 +90,13 @@ func validateAudience(boundAudiences, audClaim []string, strict bool) error {
 
 // validateBoundClaims checks that all of the claim:value requirements in boundClaims are
 // met in allClaims.
-func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]interface{}) error {
+func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims, allClaims map[string]interface{}) error {
+	if boundClaimsType != boundClaimsTypeString && boundClaimsType != boundClaimsTypeGlob {
+		return fmt.Errorf("claim type %s is not valid", boundClaimsType)
+	}
+
+	useGlobs := boundClaimsType == boundClaimsTypeGlob
+
 	for claim, expValue := range boundClaims {
 		actValue := getClaim(logger, allClaims, claim)
 		if actValue == nil {
@@ -112,10 +119,25 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
 
 	scan:
 		for _, v := range expVals {
-			for _, av := range actVals {
-				if av == v {
-					found = true
-					break scan
+			if useGlobs {
+				vs, ok := v.(string)
+				if !ok {
+					return fmt.Errorf("received claim is not a glob string: %v", v)
+				}
+				for _, av := range actVals {
+					if avs, ok := av.(string); ok {
+						if glob.Glob(vs, avs) {
+							found = true
+							break scan
+						}
+					}
+				}
+			} else {
+				for _, av := range actVals {
+					if av == v {
+						found = true
+						break scan
+					}
 				}
 			}
 		}
--- a/claims_test.go
+++ b/claims_test.go
@@ -467,6 +467,17 @@ func TestValidateBoundClaims(t *testing.T) {
 			errExpected: false,
 		},
 		{
+			name:            "not matching glob in list",
+			boundClaimsType: "glob",
+			boundClaims: map[string]interface{}{
+				"email": []interface{}{"4*d", "42*"},
+			},
+			allClaims: map[string]interface{}{
+				"email": "43x",
+			},
+			errExpected: true,
+		},
+		{
 			name:            "non matching integer glob",
 			boundClaimsType: "glob",
 			boundClaims: map[string]interface{}{
--- a/claims_test.go
+++ b/claims_test.go
@@ -181,13 +181,15 @@ func TestValidateAudience(t *testing.T) {
 
 func TestValidateBoundClaims(t *testing.T) {
 	tests := []struct {
-		name        string
-		boundClaims map[string]interface{}
-		allClaims   map[string]interface{}
-		errExpected bool
+		name            string
+		boundClaimsType string
+		boundClaims     map[string]interface{}
+		allClaims       map[string]interface{}
+		errExpected     bool
 	}{
 		{
-			name: "valid",
+			name:            "valid",
+			boundClaimsType: "string",
 			boundClaims: map[string]interface{}{
 				"foo": "a",
 				"bar": "b",
@@ -199,7 +201,8 @@ func TestValidateBoundClaims(t *testing.T) {
 			errExpected: false,
 		},
 		{
-			name: "valid - non-string claim",
+			name:            "valid - non-string claim",
+			boundClaimsType: "string",
 			boundClaims: map[string]interface{}{
 				"foo": []interface{}{42},
 			},
@@ -209,7 +212,8 @@ func TestValidateBoundClaims(t *testing.T) {
 			errExpected: false,
 		},
 		{
-			name: "valid - boolean claim",
+			name:            "valid - boolean claim",
+			boundClaimsType: "string",
 			boundClaims: map[string]interface{}{
 				"email_verified": []interface{}{false},
 			},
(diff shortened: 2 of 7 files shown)


kalafut:
Hi. Thanks for the contribution. We have generally avoided regexes in Vault, especially in the areas dealing with authentication or authorization. The main reason is that regexes are notoriously easy to get wrong, which is this case could result in unauthorized access.

Globs, on the other hand, _are_ used in Vault and are a more user-friendly way to handle many of the same use cases. Allowing globs as bound claims would be a useful improvement if you want to revise the PR.

Some specific feedback:
* Retaining an opt-in parameter is good, though I think something like `bound_claim_format` would be more flexible. Allowed today would be `"string"`, `""` (i.e. unspecified, treated as atring), or `"blob"`. This would allow future expansion, e.g. to regexes if we found a compelling reason to use them
* This blob library is recommended as it's used elsewhere in Vault: https://github.com/ryanuber/go-glob
* Is there any reason to restrict the expected claim to a single pattern, as opposed to potentially a list of patterns as is supported today?

randomswdev:
> Hi. Thanks for the contribution. We have generally avoided regexes in Vault, especially in the areas dealing with authentication or authorization. The main reason is that regexes are notoriously easy to get wrong, which is this case could result in unauthorized access.
> 
> Globs, on the other hand, _are_ used in Vault and are a more user-friendly way to handle many of the same use cases. Allowing globs as bound claims would be a useful improvement if you want to revise the PR.
> 
> Some specific feedback:
> 
>     * Retaining an opt-in parameter is good, though I think something like `bound_claim_format` would be more flexible. Allowed today would be `"string"`, `""` (i.e. unspecified, treated as atring), or `"blob"`. This would allow future expansion, e.g. to regexes if we found a compelling reason to use them
> 
>     * This blob library is recommended as it's used elsewhere in Vault: https://github.com/ryanuber/go-glob
> 
>     * Is there any reason to restrict the expected claim to a single pattern, as opposed to potentially a list of patterns as is supported today?

I modified my pull request to address all of your comments:

*  Instead of a boolean, I added a type string. Accepted values are `string` and `glob`
*  I used the glob validation library that you suggested
*  If the claim type is `glob` it is still possible associating to a claim name a list of glob expressions

Thanks for the feedback and let me know if you think additional changes are needed.

path_role.go
```diff
@@ -105,6 +108,11 @@ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 				Type:        framework.TypeCommaStringSlice,
 				Description: `Comma-separated list of 'aud' claims that are valid for login; any match is sufficient`,
 			},
+			"bound_claims_type": {
+				Type:        framework.TypeString,
+				Description: `How to interpret values in the map of claims/values (which must match for login): allowed values are string or glob`,
```
kalafut:
```suggestion
				Description: `How to interpret values in the map of claims/values (which must match for login): allowed values are 'string' or 'glob'`,
```

claims.go
```diff
@@ -89,7 +90,13 @@ func validateAudience(boundAudiences, audClaim []string, strict bool) error {
 
 // validateBoundClaims checks that all of the claim:value requirements in boundClaims are
 // met in allClaims.
-func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]interface{}) error {
+func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims, allClaims map[string]interface{}) error {
+	if boundClaimsType != boundClaimsTypeString && boundClaimsType != boundClaimsTypeGlob {
```
kalafut: Since we're checking for a valid type when the role is created, I think checking again here in an internal function is probably unnecessary.

path_role.go
```diff
@@ -426,8 +436,31 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		role.VerboseOIDCLogging = verboseOIDCLoggingRaw.(bool)
 	}
 
+	boundClaimsType := data.Get("bound_claims_type").(string)
+	if boundClaimsType != boundClaimsTypeString && boundClaimsType != boundClaimsTypeGlob {
```
kalafut: Suggested refactoring:

```
switch boundClaimsType {
    case boundClaimsTypeString, boundClaimsTypeGlob:
        role.BoundClaimsType = boundClaimsType
    default:
        return logical.ErrorResponse("invalid 'bound_claims_type': %s", boundClaimsType), nil
}
```

path_role.go
```diff
@@ -426,8 +436,31 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
 	if boundClaimsRaw, ok := data.GetOk("bound_claims"); ok {
 		role.BoundClaims = boundClaimsRaw.(map[string]interface{})
+
+		if boundClaimsType == boundClaimsTypeGlob {
+			// Check that the claims are all strings
+			for _, claimValues := range role.BoundClaims {
+				claimsValuesList, ok := normalizeList(claimValues)
+
+				if !ok {
+					return logical.ErrorResponse("received claim is not a string or list: %v", claimValues), nil
```
kalafut:
```suggestion
					return logical.ErrorResponse("claim is not a string or list: %v", claimValues), nil
```

path_role.go
```diff
@@ -426,8 +436,31 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
+				}
+
+				for _, claimValue := range claimsValuesList {
+					if _, ok := claimValue.(string); !ok {
+						return logical.ErrorResponse("received claim is not a string: %v", claimValue), nil
```
kalafut:
```suggestion
						return logical.ErrorResponse("claim is not a string: %v", claimValue), nil
```

kalafut:
Add upgrade login to the internal role read function (`b.role`), ala: 

```
if boundClaimsType == "" {
    boundClaimsType = boundClaimsTypeString
}
```

claims_test.go
```diff
@@ -382,7 +400,8 @@ func TestValidateBoundClaims(t *testing.T) {
 		},
 
 		{
-			name: "invalid received claim expected value",
+			name:            "invalid received claim expected value",
+			boundClaimsType: "string",
 			boundClaims: map[string]interface{}{
 				"email": "d",
 			},
```
kalafut: Include a test case with `boundClaimsType == ""` since that's what all existing roles will be read as.

kalafut:
@randomswdev The PR is looking really good. Just a few small comments, plus the need to handle existing roles.

path_role.go
randomswdev: Implemented the change in a new commit (to simply reviewing it)

claims.go
randomswdev: Removed the check in the new commit (to simply reviewing it)

path_role.go
randomswdev: Implemented the change in the new commit (to simply reviewing it)

path_role.go
randomswdev: Removed the check in the new commit (to simply reviewing it)

path_role.go
randomswdev: Removed the check in the new commit (to simply reviewing it)

randomswdev:
> 
> 
> Add upgrade logic to the internal role read function (`b.role`), ala:
> 
> ```
> if boundClaimsType == "" {
>     boundClaimsType = boundClaimsTypeString
> }
> ```

Added the suggested logic in the new commit (to simply reviewing it)

claims_test.go
randomswdev: Extedned an existing test case by removing the `bound_claims_type` entry from the json representation of a role, saving the role in storage and reading it back (same testing paatern as for the role_type attribute).
I completely missed the "upgrade" path, i.e. reading from storage an existing role. Thanks for catching it.

randomswdev:
I should have addressed all the review comments. I commited the changes in a separate commit to simplify further reviewing them.
Let me know if additional changes are needed.

path_role_test.go
```diff
@@ -670,6 +673,26 @@ func TestPath_Read(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err = req.Storage.Put(context.Background(), entry); err != nil {
+		t.Fatal(err)
+	}
+
+	// Run read test for "upgrade" case. The legacy role is not changed in storage, but
+	// reads will populate the `bound_claims_type` with "jwt".
```
kalafut:
```suggestion
	// reads will populate the `bound_claims_type` with "string".
```

claims.go
```diff
@@ -112,10 +115,25 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
...
 	scan:
 		for _, v := range expVals {
-			for _, av := range actVals {
-				if av == v {
-					found = true
-					break scan
+			if useGlobs {
+				vs, ok := v.(string)
+				if !ok {
+					return fmt.Errorf("received claim is not a glob string: %v", v)
```
kalafut:
```suggestion
					return fmt.Errorf("received claim is not a string: %v", v)
```

claims.go
```diff
@@ -112,10 +115,25 @@ func validateBoundClaims(logger log.Logger, boundClaims, allClaims map[string]in
 
 	scan:
 		for _, v := range expVals {
-			for _, av := range actVals {
-				if av == v {
-					found = true
-					break scan
+			if useGlobs {
+				vs, ok := v.(string)
+				if !ok {
```
kalafut: We've already verified this at role update/create time, right?  I think `vs := v.(string)` is appropriate. We'd panic if not string, but that's OK since the role code shouldn't never let non-strings become expected values.

path_role_test.go
randomswdev: Addressed in the third commit, sorry for the mistake.

claims.go
randomswdev: Addressed in the third commit.

path_role.go
```diff
@@ -426,8 +440,33 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		role.VerboseOIDCLogging = verboseOIDCLoggingRaw.(bool)
 	}
 
+	boundClaimsType := data.Get("bound_claims_type").(string)
+	switch boundClaimsType {
+	case boundClaimsTypeString, boundClaimsTypeGlob:
+		role.BoundClaimsType = boundClaimsType
+	default:
+		return logical.ErrorResponse("invalid 'bound_claims_type': %s", boundClaimsType), nil
```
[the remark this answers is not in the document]
randomswdev: Added three new test cases in the 4th commit.

claims_test.go
```diff
@@ -391,9 +410,97 @@ func TestValidateBoundClaims(t *testing.T) {
...
+			boundClaimsType: "glob",
+			boundClaims: map[string]interface{}{
+				"email": "*2",
+			},
+			allClaims: map[string]interface{}{
+				"email": "42x",
+			},
+			errExpected: true,
+		},
+		{
```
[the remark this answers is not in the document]
randomswdev: Added the test case in the 4th commit.

kalafut:
@randomswdev  Thanks!

PR #91, opened 2019-12-11
Reduce nesting in validateBoundClaims

tyrannosaurus-becks:
This PR should only be reviewed and merged after #89 is merged. It piggy-backs on that code, but rewrites `validateBoundClaims` to reduce nesting and increase maintainability. Didn't want to put it onto that PR author to worry about. After that PR is merged, this diff will be much smaller.


kalafut:
@tyrannosaurus-becks #89 is merged, which causes a couple conflicts here.

PR #92, opened 2019-12-12
Bump dependencies

andrejvanderzee:
Bump dependencies to use latest vault/api.


kalafut:
Hi. I've just bumped master to something more recent.

PR #94, opened 2019-12-13
Move tests into sub tests

pcman312:
Moved a bunch of tests in large test functions into their own sub tests using `t.Run`. This helps prevent test bleedover effects, makes the test output easier to read, and makes the tests easier to debug.

### DIFF
--- a/path_role_test.go
+++ b/path_role_test.go
@@ -38,514 +38,55 @@ func getBackend(t *testing.T) (logical.Backend, logical.Storage) {
 }
 
 func TestPath_Create(t *testing.T) {
-	b, storage := getBackend(t)
+	t.Run("", func(t *testing.T) {
+		b, storage := getBackend(t)
 
-	data := map[string]interface{}{
-		"role_type":       "jwt",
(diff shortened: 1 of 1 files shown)


path_role_test.go
```diff
@@ -38,624 +38,699 @@ func getBackend(t *testing.T) (logical.Backend, logical.Storage) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
-	if resp != nil && !resp.IsError() {
-		t.Fatalf("expected error")
-	}
-	if !strings.HasPrefix(resp.Error().Error(), "must have at least one bound constraint") {
-		t.Fatalf("unexpected err: %v", resp)
-	}
+	t.Run("", func(t *testing.T) {
```
tyrannosaurus-becks:
```suggestion
	t.Run("sunny path", func(t *testing.T) {
```

tyrannosaurus-becks: Would that be a good name for this one? Feel free to change it if that doesn't really fit.

pcman312: How about `"happy path"`?

PR #96, opened 2020-01-30
Support configuring response_mode

kalafut:
Allows configuring the `response_mode` for OIDC requests to "form_post" (the default is and will continue to be "query"). Some services, such as ADFS, require form_post in order to access certain features.


path_config.go
```diff
@@ -227,6 +233,12 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		}
 	}
 
+	switch config.OIDCResponseMode {
+	case "", "query", "form_post":
```
tyrannosaurus-becks: Not essential, but it might be worth considering using an enum for these. I use enums to be more defensive because it reduces the possibility of typos and makes it clearer what is and isn't supported. However, on the downside, sometimes an enum can add up to _more_ code. So definitely a judgement call.

PR #98, opened 2020-02-13
Support form_post mode

kalafut:
[form_post](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) is not a commonly used mode but is needed for certain
applications. Adding support for it impacted both the CLI and UI handlers,
as well as Vault (https://github.com/hashicorp/vault/pull/8325).

In addition to allowing configuration of `response_mode=form_post`, the
operator can now also choose to configure `response_types`, which today
(i.e. the released version) is always "code". This PR allows adding the `id_token`
type as well. This is only supported if form_post is enabled.

A curious aspect of this PR is to respond with some Javascript that
writes to localStorage. This is a effectively doing what the UI does
after a redirect, and allows the form_post support without modifying the
UI code at all. There are cleaner options, but they would require a
significant update to the UI.


kalafut:
Additional form_post specific tests are to be added, but I wanted to get this out ASAP.

path_oidc.go
```diff
@@ -106,6 +153,22 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
...
+		return logical.ErrorResponse(errLoginFailed + " Expired or missing OAuth state."), nil
+	}
+
+	roleName := state.rolename
+	role, err := b.role(ctx, req.Storage, roleName)
+	if err != nil {
+		return nil, err
+	}
+	if role == nil {
+		return logical.ErrorResponse(errLoginFailed + " Role could not be found"), nil
```
jasonodonnell: I actually asked core folks today if we should return `nil, nil` or an error response.  Seems like `nil, nil` was the preferred way.  Wondering if there's a special reason why we're returning a response here?

path_oidc.go
```diff
@@ -77,25 +92,57 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
...
 	}
-	if role == nil {
-		return logical.ErrorResponse(errLoginFailed + " Role could not be found"), nil
+
+	return resp, nil
+}
+
+func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	// Because the state is cached, don't process OIDC logins on perf standbys
+	if b.System().ReplicationState().HasState(consts.ReplicationPerformanceStandby) {
```
jasonodonnell: As discussed elsewhere, path operations could replace this.

path_oidc.go
kalafut: Mainly so the UI can display something more helpful/specific. The OIDC flow has more steps and things that could fail or be misconfigured, so surfacing detailed errors was a goal.

image

path_config.go
```diff
@@ -41,6 +49,14 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 					Sensitive: true,
 				},
 			},
+			"oidc_response_mode": {
+				Type:        framework.TypeString,
```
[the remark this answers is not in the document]
kalafut: That is often used for something like a role name that is user-chosen, and we'd want "Admin" and "admin" to normalize to the same thing. In this case the parameter is defined in the OAuth RFC as "form_post", and we should complain when they try to configure "Form_post".

xntrik:
Useful references for OpenID Connect and these new modes:
- https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html
- https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations

And the core OIDC spec: https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth

cli.go
```diff
@@ -132,6 +112,52 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
+		// Pull any parameters from either the body or query parameters.
+		// FormValue prioritizes body values, if found.
+		data := map[string][]string{
+			"state":    {req.FormValue("state")},
+			"code":     {req.FormValue("code")},
+			"id_token": {req.FormValue("id_token")},
+		}
+
+		// If this is a POST, then the form_post response_mode is being used and the flow
+		// involves an extra step. First GET the data to Vault, and then issue a GET with
```
xntrik: We're actually first POSTing the data to vault, then issuing a GET

cli.go
```diff
@@ -132,6 +112,52 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
+		// the same state/code to complete the auth as normal.
+		if req.Method == http.MethodPost {
+			url := c.Address() + path.Join("/v1/auth", mount, "oidc/callback")
```
xntrik: There's no existing API Client logical functions that can submit this POST request, similar to how we use `ReadWithData` below?

cli.go
```diff
@@ -132,6 +112,52 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
+				response = errorHTML(summary, detail)
+				return
+			}
+			defer resp.Body.Close()
+
+			// An id_token will never be part of a redirect GET, so remove it here too.
+			delete(data, "id_token")
+		}
+
+		secret, err = c.Logical().ReadWithData(fmt.Sprintf("auth/%s/oidc/callback", mount), data)
```
xntrik: When we get here, which is the existing pattern, do we do anything with the `resp` data from the POST above?

path_config.go
```diff
@@ -41,6 +49,14 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
...
+				Description: "The response mode to be used in the OAuth2 request. Allowed values are 'query' and 'form_post'.",
+			},
+			"oidc_response_types": {
+				Type:        framework.TypeCommaStringSlice,
+				Description: "The response typess to request. Allowed values are 'code' and 'id_token'. Defaults to 'code'.",
```
xntrik: We also support the multiple response types yeah? https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations
So allowed values may be things like 'code id_token' 'id_token code' .. ('token id_token code' ??)

path_oidc.go
```diff
@@ -134,21 +195,33 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 		Scopes:       []string{oidc.ScopeOpenID},
 	}
 
+	var rawToken string
+	var oauth2Token *oauth2.Token
+
 	code := d.Get("code").(string)
-	if code == "" {
-		return logical.ErrorResponse(errLoginFailed + " OAuth code parameter not provided"), nil
+	if code == noCode {
```
xntrik: when would this occur? Trying to understand where the 'code' field would be submitted to this callback where it equals `no_code`

path_oidc.go
```diff
@@ -284,6 +354,15 @@ func (b *jwtAuthBackend) authURL(ctx context.Context, req *logical.Request, d *f
 		return resp, nil
 	}
 
+	// If configured for form_post, redirect directly to Vault instead of the UI,
```
xntrik: Are there circumstances where perhaps the UI is disabled, and we want to support the 'cli jwt/oidc' login flow, would this break that?

path_oidc.go
```diff
@@ -308,7 +387,33 @@ func (b *jwtAuthBackend) authURL(ctx context.Context, req *logical.Request, d *f
...
 
-	resp.Data["auth_url"] = oauth2Config.AuthCodeURL(stateID, oidc.Nonce(nonce))
+	authCodeOpts := []oauth2.AuthCodeOption{
+		oidc.Nonce(nonce),
+	}
+
+	// Add "form_post" param if requested. Note: the operator is allowed to configure "query"
+	// as well, but that is the default for the AuthCode method and needn't be explicitly added.
+	if config.OIDCResponseMode == responseModeFormPost {
+		authCodeOpts = append(authCodeOpts, oauth2.SetAuthURLParam("response_mode", "form_post"))
```
xntrik: nit Worth setting it to the const value here?

xntrik:
Also noticed that when we run the authURL callback, I'm unsure if we set an expiration on the oidcState ? (https://github.com/hashicorp/vault-plugin-auth-jwt/pull/98/files#diff-21d8bdb7e1b596220622b0be6c0b8106R424)

cli.go
kalafut: Fixed

cli.go
kalafut: Nope. The `resp` would be HTML + JS for the UI to set the localStorage we we don't care about in the CLI. The POST just accomplishes the job of filling the cache in Vault.

path_config.go
kalafut: We do support multiple types, but they're provided individually (note the type `TypeCommaStringSlice`). So `oidc_response_types=code,id_token` or even `oidc_response_types=code oidc_response_types=id_token`. Order doesn't matter as we combine the segments into a string later.

kalafut: Fixed `typess`

path_oidc.go
kalafut: When the UI POSTs to the plugin, the response HTML with the JS is configured here, which references `noCode`: https://github.com/hashicorp/vault-plugin-auth-jwt/pull/98/files#diff-21d8bdb7e1b596220622b0be6c0b8106R138 and pops it in the JS here: https://github.com/hashicorp/vault-plugin-auth-jwt/pull/98/files#diff-0e83cde1b3d66eaa8aa4e534959abebbR457.  So it's a sentinel value meaning go look in the cache state for 'code', if there is any. We can't leave that field out or even empty in the JS since the UI then complains.

path_oidc.go
kalafut: 👍

kalafut:
> Also noticed that when we run the authURL callback, I'm unsure if we set an expiration on the oidcState ? (https://github.com/hashicorp/vault-plugin-auth-jwt/pull/98/files#diff-21d8bdb7e1b596220622b0be6c0b8106R424)

The cache is configured here which references the 10 minute `oidcStateTimeout` value: https://github.com/hashicorp/vault-plugin-auth-jwt/blob/026b6c94c5bf5f507cdb3c5d6cd5da5f85149277/backend.go#L46   The SetDefault cache functions will use that as the default expiration. (The 1 minute in the cache creation call is how often a goroutine runs to clear out expired entries.)

cli.go
kalafut: Most of the API Client functions choke if they don't get a secret back. There are a couple of "Raw" flavored ones, but I found them to be a pain since they're not set up for forms. So I have to manual encode everything, construct a body reader, set correct content type, etc. All the stuff `PostForm` is already handling.

xntrik: Okies

path_config.go
xntrik: Ah okay, thanks for clarification. (Feel free to resolve)

path_oidc.go
xntrik: gotcha!

xntrik:
Ah awesome - thanks for clarifying the caching expiration.

cli.go
```diff
@@ -132,6 +112,52 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 	}
 }
 
+func callbackHandler(c *api.Client, mount string, doneCh chan<- loginResp) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		var response string
+		var secret *api.Secret
+		var err error
```
tyrannosaurus-becks: This err is never populated, should we just strip it?

path_config.go
```diff
@@ -14,10 +14,18 @@ import (
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/helper/certutil"
+	"github.com/hashicorp/vault/sdk/helper/strutil"
```
tyrannosaurus-becks: These don't parse for me at first until I run `$ go mod vendor`, and then I get a diff in vendored files. So I think we need `$ go mod vendor` run on this PR.

path_config.go
```diff
@@ -227,6 +247,22 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
...
+	}
+
+	// Validate response_mode
+	switch config.OIDCResponseMode {
+	case "", responseModeQuery:
+		if config.hasType(responseTypeIDToken) {
+			return logical.ErrorResponse("query response_mode may not be used with an id_token response_type"), nil
+		}
+	case responseModeFormPost:
+	default:
```
tyrannosaurus-becks: This can be stripped because the "" case is the default and is covered above.

path_config.go
```diff
@@ -227,6 +247,22 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
...
+		return logical.ErrorResponse("invalid response_types %v. 'code' and 'id_token' are allowed", config.OIDCResponseTypes), nil
+	}
+
+	// Validate response_mode
+	switch config.OIDCResponseMode {
+	case "", responseModeQuery:
+		if config.hasType(responseTypeIDToken) {
+			return logical.ErrorResponse("query response_mode may not be used with an id_token response_type"), nil
+		}
+	case responseModeFormPost:
```
tyrannosaurus-becks: Nit: might be nice to just add a `// Pass` or `// Success` comment on the next line down so maintainers don't think it's an oversight that there's no code following. Or alternatively, since there's only one case being checked now, could also just go to an `if` statement.

path_config.go
```diff
@@ -286,6 +322,8 @@ type jwtConfig struct {
 	OIDCDiscoveryCAPEM   string   `json:"oidc_discovery_ca_pem"`
 	OIDCClientID         string   `json:"oidc_client_id"`
 	OIDCClientSecret     string   `json:"oidc_client_secret"`
+	OIDCResponseMode     string   `json:"oidc_response_mode"`
+	OIDCResponseTypes    []string `json:"oidc_response_types`
```
tyrannosaurus-becks:
```suggestion
	OIDCResponseTypes    []string `json:"oidc_response_types"`
```

path_oidc.go
```diff
@@ -46,12 +53,25 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
...
+				},
 			},
 
 			Operations: map[logical.Operation]framework.OperationHandler{
 				logical.ReadOperation: &framework.PathOperation{
 					Callback: b.pathCallback,
 					Summary:  "Callback endpoint to complete an OIDC login.",
+
+					// state is cached so don't process OIDC logins on perf standbys
+					ForwardPerformanceStandby: true,
```
tyrannosaurus-becks: This is my first time working with this field and it's not deeply described on the path operation. The way I read it is, by default it's false, so by default it would read "ForwardPerformanceStandby: false" and the resulting behavior would be that we _would not_ forward requests to performance standby's, we'd only handle them on the leader. 

So the way I read this, setting it to true would cause it to yes, forward requests to performance standbys. However, the comment reads the opposite way.

Am I thinking of this incorrectly? Same question for line 74.

path_oidc.go
```diff
@@ -71,33 +91,58 @@ func pathOIDC(b *jwtAuthBackend) []*framework.Path {
...
-	if b.System().ReplicationState().HasState(consts.ReplicationPerformanceStandby) {
-		return nil, logical.ErrReadOnly
+	if config.OIDCResponseMode != responseModeFormPost {
+		return logical.RespondWithStatusCode(nil, req, http.StatusMethodNotAllowed)
 	}
 
-	state := b.verifyState(d.Get("state").(string))
-	if state == nil {
-		return logical.ErrorResponse(errLoginFailed + " Expired or missing OAuth state."), nil
+	stateID := d.Get("state").(string)
```
tyrannosaurus-becks: Should we be calling `verifyState` anywhere in here? Forgive my lack of familiarity with this codebase, I only visit it from time to time!

tyrannosaurus-becks: I suppose it's probably intentionally this way because `amendState` looks very similar, but just want to understand it better.

path_oidc.go
```diff
@@ -13,26 +15,31 @@ import (
...
+)
 
 // oidcState is created when an authURL is requested. The state identifier is
 // passed throughout the OAuth process.
 type oidcState struct {
 	rolename    string
 	nonce       string
 	redirectURI string
+	code        string
+	id_token    string
```
tyrannosaurus-becks: Needs to be refactored to camel case vs. snake case.

path_oidc.go
```diff
@@ -164,17 +237,19 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
...
-	} else {
-		logFunc := b.Logger().Warn
-		if strings.Contains(err.Error(), "user info endpoint is not supported") {
-			logFunc = b.Logger().Info
+	// If we have a token, attempt to fetch information from the /userinfo endpoint
+	// and merge it with the existing claims data. A failure to fetch additional information
+	// from this endpoint will not invalidate the authorization flow.
+	if oauth2Token != nil {
+		if userinfo, err := provider.UserInfo(oidcCtx, oauth2.StaticTokenSource(oauth2Token)); err == nil {
+			_ = userinfo.Claims(&allClaims)
```
tyrannosaurus-becks: Should we be dropping the error here?

path_oidc.go
```diff
@@ -379,3 +499,15 @@ func validRedirect(uri string, allowed []string) bool {
 
 	return false
 }
+
+// parseMount attempts to extract the mount path from a redirect URI.
+func parseMount(redirectURI string) string {
```
tyrannosaurus-becks: Bonus - would be nice to see an example string here.

tyrannosaurus-becks:
Getting a test failure on this branch presently:
```
--- FAIL: TestConfig_ResponseMode (0.00s)
    path_config_test.go:294: expected error, got none for "QUERY"
```

path_config.go
tyrannosaurus-becks: I'm incorrect about this. Resolving.

cli.go
tyrannosaurus-becks: This is addressed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/99 if we merge it.

path_config.go
tyrannosaurus-becks: This is addressed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/99 if we merge it.

path_config.go
tyrannosaurus-becks: This is addressed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/99 if we merge it.

path_oidc.go
tyrannosaurus-becks: This is addressed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/99 if we merge it.

path_oidc.go
tyrannosaurus-becks: Nevermind! I had a chance to go read the Vault code and answered my own question.

tyrannosaurus-becks: Should we also use `ForwardPerformanceSecondary: true` for these?

path_oidc.go
tyrannosaurus-becks: For this one, I wrote a test on a branch like this:
```
func TestParseMount(t *testing.T) {
	result := parseMount("https://example.com/v1/auth/oidc/foo")
	fmt.Println(result)
}
```
When I run that, it prints out "oidc". It looks like the mount path _can_ include any ascii characters, and "/" is one of them. (https://www.vaultproject.io/api/system/mounts.html#inlinecode-path-14.)

If I change the input to "https://example.com/v1/auth/oidc", I still get "oidc". So it definitely works in the most common use but I think we need to change the algorithm in here a little.

tyrannosaurus-becks: This is addressed in https://github.com/hashicorp/vault-plugin-auth-jwt/pull/99 if we accept it.

tyrannosaurus-becks:
This code looks great and I approve it as long as we merge the linked PR and double-check the one err we're dropping above. This can be considered approved by me even with those minor subsequent changes.

PR #99, opened 2020-02-14
Easily addressable PR feedback

tyrannosaurus-becks:
This PR incorporates the PR feedback I've given that is low-risk in addressing. The commit history cleanly describes the changes, it would be best to look one at a time.


cli.go
```diff
@@ -116,11 +116,10 @@ func callbackHandler(c *api.Client, mount string, doneCh chan<- loginResp) http.
 	return func(w http.ResponseWriter, req *http.Request) {
 		var response string
 		var secret *api.Secret
-		var err error
 
 		defer func() {
 			w.Write([]byte(response))
-			doneCh <- loginResp{secret, err}
+			doneCh <- loginResp{secret, nil}
```
kalafut: The error that was here is the one that gets set on line 150.

path_oidc.go
```diff
@@ -502,12 +502,10 @@ func validRedirect(uri string, allowed []string) bool {
 
 // parseMount attempts to extract the mount path from a redirect URI.
 func parseMount(redirectURI string) string {
-	parts := strings.Split(redirectURI, "/")
-
-	for i := 0; i+2 < len(parts); i++ {
-		if parts[i] == "v1" && parts[i+1] == "auth" {
-			return parts[i+2]
-		}
+	parts := strings.Split(redirectURI, "/v1/auth/")
```
kalafut: Nice.

kalafut: Actually...it doesn't work and I restore the old code and revised the test. We only want the immediate next element, not the rest of the path.

PR #104, opened 2020-03-12
Add optional client_nonce for OIDC logins

kalafut:
This optional nonce hardens the multi-step OIDC login process. If the
client provides a nonce at the start (auth_url) phase, it must provide
the same nonce as part of the callback.

The motivation for this is primarily CLI logins. The CLI listener is
necessarily unencrypted, and theoretically if an attacker stole the OAuth
authorization_code enroute to the CLI listener, it could post that to
Vault and gain access. The CLI leverages the new client_nonce support to
harden this flow. Only the CLI process (and Vault) will know the
client_nonce, so an attacker will be unable to complete the login even
if they manage to steal the authorization_code.

### DIFF
--- a/path_oidc_test.go
+++ b/path_oidc_test.go
@@ -750,7 +750,7 @@ func TestOIDC_Callback(t *testing.T) {
 			// set mock provider's expected code
 			s.code = "abc"
 
-			// invoke the callback, which will in to try to exchange the code
+			// invoke the callback, which will try to exchange the code
 			// with the mock provider.
 			req = &logical.Request{
 				Operation: logical.ReadOperation,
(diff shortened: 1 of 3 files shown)


path_oidc_test.go
```diff
@@ -685,6 +685,95 @@ func TestOIDC_Callback(t *testing.T) {
...
+			state := getQueryParam(t, authURL, "state")
+			nonce := getQueryParam(t, authURL, "nonce")
+
+			// set provider claims that will be returned by the mock server
+			s.customClaims = sampleClaims(nonce)
+
+			// set mock provider's expected code
+			s.code = "abc"
+
+			// invoke the callback, which will in to try to exchange the code
```
tvoran:
```suggestion
			// invoke the callback, which will try to exchange the code
```

Issue #101, opened 2020-03-06
Save refresh tokens and support obtaining new jwt access tokens from them


DrDaveD:
We would like to use vault to store the "refresh_token" that can come during an OIDC authorization flow, and be able to use that refresh token to obtain short-lived jwt access tokens without having to go through the authorization flow again.  This seems like it would fit in very well with this vault oidc/jwt plugin, but so far I see no hit on the word "refresh" anywhere in this repository.

Do you agree this is a good functionality to have in this plugin?  Is it in an existing plan?  If I were to develop the code in order to speed its availability, would you be willing to provide me with some hints if I get stuck?

Currently our science computing communities ([Open Science Grid](https://opensciencegrid.org) and the [Worldwide LHC Computing Grid](https://wlcg.web.cern.ch/)) has been focusing on another tool called [oidc-agent](https://github.com/indigo-dc/oidc-agent) which does provide that functionality but it is more oriented towards individual users and we're looking for a server-based solution that can be shared by large groups of users.

kalafut:
Hi. Can you explain a little more the flow(s) you envision? Are you proposing to use the refresh/access tokens as a means to authenticate with Vault again, or for other means, and merely capturing & returning them as a result of OIDC auth?

DrDaveD:
Merely as a means of capturing & returning them.  I envision using regular vault access tokens and Kerberos to access vault, and then using the retrieved OIDC access tokens to access other services.

More specifically I want vault to save the OIDC refresh tokens, cache the access tokens, renew the access tokens when needed (that is, when there's less than a configurable amount of life remaining in them) using the refresh tokens, and return access tokens to the user.  OIDC token issuers (optionally) return refresh tokens when the scope "offline_access" is included.

kalafut:
Thanks for the explanation. The basic roles of Vault's auth methods are authenticating to Vault and establishing identity, and the use case you've described does seem to go beyond that.  Data from the auth process is generally not returned, though in some cases portions of it are attached to identity metadata if appropriate. The data retained isn't secret, and is often descriptive properties such as the user's region, or resource group.

Have you considered running the plugin externally (i.e. as an external plugin, overriding the  version bundled in Vault)? You'd be free to customize the functionality as needed.

DrDaveD:
[...]

PR #107, opened 2020-04-02
Add ability to store refresh token into a secrets plugin

DrDaveD:
# Overview
This PR adds configuration options to store a received refresh_token into the given path in vault.  It has been tested with the [Puppet Labs oauthapp vault plugin](https://github.com/puppetlabs/vault-plugin-secrets-oauthapp) and the corresponding [pull request](https://github.com/puppetlabs/vault-plugin-secrets-oauthapp/pull/6) that accepts the refresh token.

These options make it possible for a user to authenticate with vault once and from then on be able to get access tokens from the OIDC provider via the secrets plugin.  This is needed by the High Energy Physics science community as they migrate their global computing from X.509 certificates to OIDC tokens.

# Design of Change
The role configuration option `refresh_store_path` can be set to specify a vault path to write into.  The path may contain any claim names from the id_token surrounded by double curly brackets (for example "{{email}}", which are replaced by the value of the claim.  The role configuration option `refresh_store_cred` should contain a vault token with sufficient privileges to write to the path, or the token set in the environment variable VAULT_TOKEN.

Addresses issue #101.

# Contributor Checklist
I will write the docs if this PR is otherwise deemed acceptable.

[  ] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[x] Backwards compatible

### DIFF
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -337,18 +337,18 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 		}
 		storePath := role.RefreshStorePath
 		for {
-			/* replace claims in storePath of form {claim} with its value */
-			start := strings.Index(storePath, "{")
+			/* replace claims in storePath of form {{claim}} with its value */
+			start := strings.Index(storePath, "{{")
 			if start == -1 {
 				break
 			}
-			end := strings.Index(storePath, "}")
+			end := strings.Index(storePath, "}}")
 			if end < start {
 				return logical.ErrorResponse("mismatched brackets in refresh_store_path %s", role.RefreshStorePath), nil
 			}
-			claim := storePath[start+1:end]
+			claim := storePath[start+2:end]
 			if val, ok := allClaims[claim]; ok {
-				storePath = strings.ReplaceAll(storePath, "{" + claim + "}", val.(string))
+				storePath = strings.ReplaceAll(storePath, "{{" + claim + "}}", val.(string))
 			} else {
 				return logical.ErrorResponse("no claim %s found for refresh_store_path %s", claim, role.RefreshStorePath), nil
 			}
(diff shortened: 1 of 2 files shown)


DrDaveD:
I changed the single curly brackets around the claim in refresh_store_path to double curly brackets to match the usage in vault policies, and edited the description to reflect the change.

DrDaveD:
Is there any reaction to this by the plugin owners?

DrDaveD:
@kalafut @jefferai Do you have any thoughts on this?

kalafut:
@DrDaveD Thank you for your work on this and for the submission to Puppet's oauth secrets plugin (which I learned about through this issue). To date we've not allowed the pattern of plugins calling other Vault plugins or endpoints. More generally, the storage of a refresh token during Vault authentication is a fairly niche use case that we'll not be supporting at this time.

Maintaining a fork of the JWT/OIDC repo is the recommended path in this case. In #101 you had questions about replacing the internal JWT/OIDC mount with an external plugin, and that is definitely possible. If you have further questions about that process, the best venue would be to post to https://discuss.hashicorp.com/c/vault.

jmls:
dang, looks like I'm a niche case as well ... would have found this useful. @DrDaveD , could I ask how you distribute your custom version of this plugin ? Do you build a new vault image or deregister the official plugin and register yours the first time you start vault ? I have a change that I'd like to make (#109 ) and would want to include this PR intpo my version.

I just am not sure about the whole "replace a plugin with another" scenario ;)

DrDaveD:
@jmls I don't distribute this yet but if needed my plan would be to distribute the modified plugin only, and use it to replace the standard one.  That's what I am doing for testing.

Note that I have also now proposed a simpler alternative to this PR in #113 which just makes the refresh token available to the client in metadata.  If that's sufficient for your needs, I suggest that you comment on that PR to express your support.

PR #113, opened 2020-05-08
Add refresh_field option to put refresh token in metadata

DrDaveD:
# Overview
This PR adds a configuration option refresh_field to return the refresh token in metadata under the name given by the option.

This options makes it possible for a user to authenticate with vault once and with some help from a client application store the token back into a vault secrets plugin like the [Puppet labs oauthapp](https://github.com/puppetlabs/vault-plugin-secrets-oauthapp).  From then on access tokens can be read from the secrets plugin.  This is needed by the High Energy Physics science community including the thousands of members of the collaborations running experiments at the Large Hadron Colider.

# Design of Change

The new option simply puts the refresh token into the returned metadata in a similar way to the claims_mapping option.

This is another way of addressing issue #101 that is simpler than the rejected pull request #107.

# Contributor Checklist
I will write the docs if this PR is otherwise deemed acceptable.

[x] Backwards compatible

### DIFF
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -215,6 +215,7 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 	}
 
 	var rawToken string
+	var refreshToken string
 	var oauth2Token *oauth2.Token
 
 	code := d.Get("code").(string)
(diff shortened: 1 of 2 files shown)


kalafut:
Hi @DrDaveD. This seems to work fine, though I wonder if the UX could be changed slightly. One option would be to have a parameter like `oauth2_metadata`, which would be a list accepting one or more of {`access_token`, `refresh_token`}. The selected data would be added to token metadata under fixed key names, e.g. `oauth2_access_token`. This would match some existing metadata config parameters in other parts of Vault a bit better. Thoughts?

DrDaveD:
That's a great idea.  I made a new PR #119 with that implementation and will close this PR.

PR #115, opened 2020-05-16
Use a mux

42wim:
Use a new mux instead of the default one so that the `Auth`  method can be used in long-running programs.
This fixes the `panic: http: multiple registrations for /oidc/callback` when running `Auth` again.


kalafut:
Hi. This looks reasonable, though can you elaborate on the use case, which sounds interesting?

42wim:
I'm writing a windows ssh agent that automatically uses vault to sign ssh public keys and then loads those ssh certificates. The necessary vault token is gotten by the `Auth` function.

Users can then chose to renew those certificates in the interface, this again fires the same oidc login flow to get a new token to sign the new generated ssh pub/priv keys

42wim:
After testing, some more changes were needed:

- Disable keepalives so that the `Serve` function can exit ASAP
- We need an extra buffer in `doneCh` because when everythings works like expected we got the secret on the `doneCh` in the select. But when the `Serve` functions exits it'll also send on the `doneCh` that now doesn't has a receiver anymore which results in a goroutine leak.

https://github.com/hashicorp/vault-plugin-auth-jwt/blob/175b36bab10d932d581b01db8e94dd78bd23a49e/cli.go#L106-108

42wim:
Can I do anything to get this merged? Any issues with this PR?

42wim:
Any maintainer that can take a look?

cli.go
```diff
@@ -121,8 +121,12 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
...
-	doneCh := make(chan loginResp)
-	http.HandleFunc("/oidc/callback", callbackHandler(c, mount, clientNonce, doneCh))
+	doneCh := make(chan loginResp, 2)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/oidc/callback", callbackHandler(c, mount, clientNonce, doneCh))
+	srv := &http.Server{Handler: mux}
+	srv.SetKeepAlivesEnabled(false)
+	defer srv.Close()
 
 	listener, err := net.Listen("tcp", listenAddress+":"+port)
```
austingebauer, line 131: Trying to understand your use case a bit more -- Are you doing something to avoid port collision? I have to imagine so if you were hitting the "multiple registrations" issue.

austingebauer (commented):
Hi @42wim. I'm going to have a look and test this out. I understand that it's helpful for how you're planning to use this. I'd like to understand if there are any undesired changes for _most_ using the regular CLI workflow. Thanks!

cli.go
42wim: This is used in a long running process, like mentioned above:

"I'm writing a windows ssh agent that automatically uses vault to sign ssh public keys and then loads those ssh certificates. The necessary vault token is gotten by the Auth function.

Users can then chose to renew those certificates in the interface, this again fires the same oidc login flow to get a new token to sign the new generated ssh pub/priv keys"

I've made a POC about the issue, see https://gist.github.com/42wim/e97d4dbf9d61ccfe436f535660d9c069 
Running this with the current code, will give you `panic: http: multiple registrations for /oidc/callback`

Because it cannot re-register the same pattern in the default mux
https://github.com/golang/go/blob/a031f4ef83edc132d5f49382bfef491161de2476/src/net/http/server.go#L2529-L2531

This is not an issue in the CLI where you only run this once and it exits, but when running multiple times in a daemon/client app it is :)

danielskowronski: This behaviour 100% matches what's seen in the official Terraform provider for Vault - https://github.com/hashicorp/terraform-provider-vault/issues/2131

I think maintainers should definitely look at it, since it affects their own product :)

torvitas:
We are using this patch in our test setup where we run multiple tests in parallel. Is there any chance this is going to be merged?

Issue #103, opened 2020-03-11
Support OAuth 2.0 device flow


DrDaveD:
I suggest that this plugin optionally support the [OAuth 2.0 device flow](https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15) for authentication.  This avoids needing to have a callback to a web server on the client's machine and is much easier to manage from the command line.

DrDaveD:
This would probably be based on this [unmerged golang oauth2 pull request](https://github.com/golang/oauth2/pull/356) or this [add-on package](https://github.com/rjw57/oauth2device).

kalafut:
Hi. Thanks for the references. Support for Device Flow is definitely of interest, especially for remote login support. It hasn’t been investigated in detail and there isn’t an ETA at this time, but it’s something that will be looked into.

DrDaveD:
I am beginning to work on a pull request for this.

kalafut:
Hi @DrDaveD. Can you summarize the approach you plan to take?

sidewinder12s:
Was able to work around this by adding this flag when SSHing into a remote machine:

```
ssh -L 8250:localhost:8250 <address>
```

And then clicking on the Auth URL when it was presented.

tsipinakis:
4 years later pinging this issue again as it's a much desired feature for us - it breaks a lot of workflows to not be able to log in easily on SSH (if you forward the port then you can't log in locally).

DrDaveD:
The port of #131 is now available in an [openbao 2.1.0 beta release](https://github.com/openbao/openbao/releases/tag/v2.1.0-beta20241114.3) via  openbao/openbao#319.

PR #122, opened 2020-07-06
Add support for OIDC device flow

DrDaveD:
# Overview
This PR adds support for the OIDC device flow.  This is a much simpler authorization flow for command line usage, because it does not require an http callback to the client command.   Requested and discussed in #103.

# Design of Change
To enable device flow, first a new config setting oidc_device_auth_url must be set containing the full URL to the OIDC provider's device endpoint.  This isn't ideal, but since golang/oauth2 does not yet support device flow, we don't have access to the device endpoint under the discovery URL.  An alternative that could be done is to implement our own discovery of that endpoint.

Once oidc_device_auth_url is set, an individual role selects device flow by setting role_type="oidcdevice".   Other roles can still use code flow with role_type="oidc" (which is still the default).

A small implementation of the oauth2 device flow that someone wrote in 2014 was copied here, and updated to be compliant with [RFC8628](https://tools.ietf.org/html/rfc8628).

Next, the `oidc/auth_url` API is modified to return the different "auth_url" that the user has to visit to authorize in a web browser.  It will be a complete url if provided, otherwise there will be a "user_code" that has to also be shown to a user to separately enter into the web browser.  There will always be a "device_code" to be passed back to the next API, and optionally there can be a default "interval" in seconds between polls for responses.

When a "device_code" is present, then next a new API `oidc/device_wait` has to be called instead of the `oidc/callback` used by the code flow.  It needs to be given the "device_code", the "role", and optionally an "interval".   This will wait for the user to respond, checking with the OIDC provider every interval seconds, and in the end return the same data that `oidc/callback` returns.

Finally, the cli is modified to work with both the old API and the new one.  There's no option to override the interval, but it could be added if requested.  The default is 5 seconds if not set by the OIDC provider, which is what the RFC recommends.

[ ] [Issue #103](https://github.com/hashicorp/vault/issues/103)

# Contributor Checklist
I will modify the relevant docs if this PR is acceptable
[x] Backwards compatible


DrDaveD:
I decided to address the issue of having to specify the device_auth_url in the role, even though the oidc_discovery_url is in config.  I moved device_auth_url in role to oidc_device_auth_url in config, and added a third role_type of 'oidcdevice'.  I will now edit the description to match the latest commit.

DrDaveD:
I now prefer my alternate implementation in #131.  I'm leaving this open for now though, because that one depends on whether or not my direct code flow callback pr #130 is accepted.

DrDaveD:
Superceded by #131

PR #137, opened 2020-10-06
Improve authorization error

ramonberrutti:
# Overview

When using a custom allowed_redirect_uris for specific roles
we don't show a relationship between the role and the
redirect_uri when an authorization error occurred,
frustrating the user thinking that the problem is in
the role and not in the allowed_redirect_uris flag.

This error is common occurred with the CLI when the administrator doesn't add an allowed localhost path.


kalafut (approved):
This LGTM and is a good idea. Thanks!

PR #141, opened 2020-10-22
Remove provider_config DisplayAttrs.Value

tvoran:
# Overview
Remove provider_config DisplayAttrs.Value, since that is treated as the default by the UI, and the default should
be empty. This was causing issues with the UI.

# Contributor Checklist
- [x] Backwards compatible


calvn (approved):

calvn:
Thanks for addressing this!

PR #143, opened 2020-10-29
Backport 1.6.x: Remove provider_config DisplayAttrs.Value (#141)

tvoran:
Backporting https://github.com/hashicorp/vault-plugin-auth-jwt/pull/141 to 1.6 branch.


calvn:
I'm surprised that release/vault-1.6.x doesn't already contain this... We might have created the branch from an older commit and not from master's latest commit.

PR #154, opened 2021-01-25
Strip signature from debug-logged token

kalafut:
The intent of this debug logging is to help understand the structure of the received token. The signature isn't needed for this so let's log the minimum required information.

### DIFF
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -251,7 +251,7 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 			loggedToken = fmt.Sprintf("%s.%s.xxxxxxxxxxx", parts[0], parts[1])
 		}
 
-		b.Logger().Debug("OIDC provider response", "ID token", loggedToken)
+		b.Logger().Debug("OIDC provider response", "id_token", loggedToken)
 	}
 
 	// Parse and verify ID Token payload.
--- a/path_oidc.go
+++ b/path_oidc.go
@@ -243,7 +243,15 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 	}
 
 	if role.VerboseOIDCLogging {
-		b.Logger().Debug("OIDC provider response", "ID token", rawToken)
+		loggedToken := "invalid token format"
+
+		parts := strings.Split(rawToken, ".")
+		if len(parts) == 3 {
+			// strip signature from logged token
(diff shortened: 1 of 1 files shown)


austingebauer (approved):
👍

path_oidc.go
```diff
@@ -243,7 +243,15 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
...
-		b.Logger().Debug("OIDC provider response", "ID token", rawToken)
+		loggedToken := "invalid token format"
+
+		parts := strings.Split(rawToken, ".")
+		if len(parts) == 3 {
+			// strip signature from logged token
+			loggedToken = fmt.Sprintf("%s.%s.xxxxxxxxxxx", parts[0], parts[1])
+		}
+
+		b.Logger().Debug("OIDC provider response", "ID token", loggedToken)
```
calvn, line 254: Nit:
```suggestion
		b.Logger().Debug("OIDC provider response", "id_token", loggedToken)
```

path_oidc.go
```diff
@@ -243,7 +243,15 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
 	}
 
 	if role.VerboseOIDCLogging {
-		b.Logger().Debug("OIDC provider response", "ID token", rawToken)
+		loggedToken := "invalid token format"
+
+		parts := strings.Split(rawToken, ".")
+		if len(parts) == 3 {
+			// strip signature from logged token
+			loggedToken = fmt.Sprintf("%s.%s.xxxxxxxxxxx", parts[0], parts[1])
```
calvn, line 251: I wish this could be represented as a valid JWT token so that it could be parsed, but this probably the cheap way to do it without having to construct the signature from scratch using an empty secret value.

calvn (approved):

path_oidc.go
kalafut: The popular parsers still show the claims info and structure, just with a warning about the signature. And you can always just grab a segment (probably the middle one) and pass it in to a base64 decoder directly.

calvn: Retracting my initial comments after giving it some more thought. Making it look more like the actual JWT might end up being worse UX since the user can't tell if that's the actual JWT token or not (they might assume it is but it's not) without digging into the specifics of how this value is constructed.

xntrik (approved):

Issue #135, opened 2020-08-26
JWKS URL changes aren't reflected immediately


lenaing:
Hello,

I encounter the following issue:

When the JWKS URL of an JWT Auth is updated in an high availability setup, the new JWKS URL is called by Vault at configuration time for validation but the old URL is still in use until the Vault switches of master.

Setup : Non-enterprise Vault version **1.5.2** (Also tested on 1.4.2), 3 servers, PostgreSQL HA and Storage Backend

How to reproduce:

* I spun up a test Vault HA Cluster with two small Python webservers (lets call them URL **A** & **B**) with both the same JWKSet with expired JWKeys. This force Vault to request JWKS each time and, I hoped, to avoid key caching.

* I configure the JWT endpoint in Vault through the UI with URL **A**.
* Vault requests URL **A** for initial configuration validation.
* Then I start to auth with a sample JWT against this Vault like this:

```
vault write auth/jwt/login role=test-role jwt="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
```

* Vault always requests URL **A** upon each authent request (so no cache ?, weird, sometime it seems more than what is specified [here](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/5aa3d0e64032eaf92497a9084df4a1ac0e114b63/vendor/github.com/coreos/go-oidc/jwks.go#L24) ).

* At this point, I change the JWT endpoint's JWKS URL to URL **B**.
* Vault requests URL **B** for configuration validation.
* When I try again to authenticate against Vault, it **still requests URL A**, even after 10 minutes.

* Then I issue a `vault operator step-down` to change the Vault master.
* Upon new authent request, Vault now calls correctly URL **B**.

URL **A** logs:
```
xx.xx.xx.97 - - [26/Aug/2020 10:37:26] "GET /test-jwks.json HTTP/1.1" 200 - Authent request
xx.xx.xx.97 - - [26/Aug/2020 10:39:12] "GET /test-jwks.json HTTP/1.1" 200 - Authent request
[...]

PR #155, opened 2021-01-25
Use the cap/jwt library for JWT based authentication

austingebauer:
## Description

This PR transitions the plugin to use the [cap/jwt](https://github.com/hashicorp/cap) library for JWT based authentication. The code in the cap/jwt library is directly inspired by the code that is in this plugin.

Fixes: https://github.com/hashicorp/vault-plugin-auth-jwt/issues/135

## Testing

All existing tests are passing against the changes in this PR. I've also manually tested the changes in this PR by running the plugin configured for JWT validation with `oidc_discovery_url`, `jwks_url`, and `jwt_validation_pubkeys`.


backend.go
```diff
@@ -129,27 +131,42 @@ func (b *jwtAuthBackend) getProvider(config *jwtConfig) (*oidc.Provider, error)
 	return provider, nil
 }
 
-// getKeySet returns a new JWKS KeySet based on the provided config.
-func (b *jwtAuthBackend) getKeySet(config *jwtConfig) (oidc.KeySet, error) {
+// getJWTValidator returns a new JWT validator based on the provided config.
+func (b *jwtAuthBackend) getJWTValidator(config *jwtConfig) (*jwt.Validator, error) {
```
calvn, line 135: Nit: Maybe we should take this chance to rename the func to `jwtValidator()`. Getters shouldn't need the `get*` prefix on their func names.

path_login.go
```diff
@@ -88,129 +86,33 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
-		}
+	// Validate the JWT by verifying it's signature and asserting expected claims values
+	allClaims, err := validator.Validate(ctx, token, expected)
+	if err != nil {
+		return logical.ErrorResponse(errwrap.Wrapf("error validating token: {{err}}", err).Error()), nil
+	}
 
-	default:
-		return nil, errors.New("unhandled case during login")
+	// If the role has no bound audiences, then return an error if any audience in the JWT claim
```
calvn, line 112: The comment here seems to be incomplete.

path_config.go
```diff
@@ -245,13 +245,11 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 		return logical.ErrorResponse("'oidc_discovery_url' must be set for OIDC"), nil
 
 	case config.JWKSURL != "":
-		ctx, err := b.createCAContext(context.Background(), config.JWKSCAPEM)
+		keyset, err := jwt.NewJSONWebKeySet(ctx, config.JWKSURL, config.JWKSCAPEM)
```
calvn, line 248: It looks like the `keyset.VerifySignature(ctx, testJWT)` will no longer use a context with the config's CA.

I wonder if the `VerifySignature` call in here should be storing and using the `caCtx` that got instantiated in `NewJSONWebKeySet`.

backend.go
austingebauer: Yep, might as well do that now. Done in 722a769.

path_login.go
austingebauer: I've updated the comment in 1ff98b6. Let me know if that makes more sense!

path_login.go
```diff
@@ -88,129 +86,33 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+		ExpirationLeeway:  role.ExpirationLeeway,
+		ClockSkewLeeway:   role.ClockSkewLeeway,
+	}
 
-	case configType == OIDCDiscovery:
-		allClaims, err = b.verifyOIDCToken(ctx, config, role, token)
-		if err != nil {
-			return logical.ErrorResponse(err.Error()), nil
-		}
+	// Validate the JWT by verifying it's signature and asserting expected claims values
```
calvn, line 106:
```suggestion
	// Validate the JWT by verifying its signature and asserting expected claims values
```

path_login_test.go
```diff
@@ -64,6 +64,7 @@ func setupBackend(t *testing.T, cfg testConfig) (closeableBackend, logical.Stora
 			data = map[string]interface{}{
 				"bound_issuer":           "https://team-vault.auth0.com/",
 				"jwt_validation_pubkeys": ecdsaPubKey,
+				"jwt_supported_algs":     []string{string(jose.ES256)},
```
calvn, line 67: Should we be testing with `cap/jwt` (i.e. `jwt.ES256`) values instead?

path_config.go
austingebauer: The `keyset.VerifySignature()` call below actually does end up using the context with the config's CA that's created in cap/jwt (see [keyset.go#L114](https://github.com/hashicorp/cap/blob/main/jwt/keyset.go#L114)). That's because the underlying `remoteKeySet` stores the context (see [jwks.go#L49](https://github.com/coreos/go-oidc/blob/v2/jwks.go#L49)) and uses the associated client with the config's CA to make the HTTP request for the JWKS (see [jwks.go#L198](https://github.com/coreos/go-oidc/blob/v2/jwks.go#L198)). The context passed to `keyset.VerifySignature()` isn't used for the JWKS request.

calvn: Ah yep, you're right! Thanks for the pointer on `updateKeys()`.

path_login_test.go
austingebauer: Ya, can use the values from `cap/jwt` here. I've done that in 22e3e09. There was a package name collision with `square/go-jose.v2/jwt` though, so I changed its identifier. Lmk your thoughts!

calvn (approved):
Changes looks good, but you might want to get Jim's 👀 's on this one since he's more familiar with the codebase than I am.

path_login.go
```diff
@@ -88,129 +86,34 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
-			cksLeeway = jwt.DefaultLeeway
-		}
-
-		if err := claims.ValidateWithLeeway(expected, cksLeeway); err != nil {
-			return logical.ErrorResponse(errwrap.Wrapf("error validating claims: {{err}}", err).Error()), nil
-		}
+	// Get the JWT validator based on the configured auth type
+	validator, err := b.jwtValidator(config)
+	if err != nil {
+		return logical.ErrorResponse(errwrap.Wrapf("error configuring token validator: {{err}}", err).Error()), nil
```
kalafut, line 92: I know the pattern predate this text, but I'm not seeing the need for wrapping since the resulting parameter is just a string.  Is this the same?
```
	return logical.ErrorResponse("error configuring token validator: %s", err.Error()), nil
```

path_login.go
```diff
@@ -88,129 +86,34 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
 
-	case configType == OIDCDiscovery:
-		allClaims, err = b.verifyOIDCToken(ctx, config, role, token)
-		if err != nil {
-			return logical.ErrorResponse(err.Error()), nil
-		}
+	// Validate the JWT by verifying its signature and asserting expected claims values
+	allClaims, err := validator.Validate(ctx, token, expected)
+	if err != nil {
+		return logical.ErrorResponse(errwrap.Wrapf("error validating token: {{err}}", err).Error()), nil
```
kalafut, line 109: Same as previous comment

path_login.go
austingebauer: Yep, the wrapping is not needed. The formatted string you provided results in the same error response. Done in a9828c5.

path_login.go
austingebauer: Done in a9828c5.

PR #158, opened 2021-02-01
Use the cap/oidc library for OIDC based authentication

austingebauer:
## Description

This PR transitions the plugin to use the [cap/oidc](https://github.com/hashicorp/cap) library for OIDC based authentication.

This PR also includes changes to force the Azure-specific handling of the [groups overage claim](https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#groups-overage-claim) to use the Microsoft Graph API instead of the [deprecated](https://developer.microsoft.com/en-us/office/blogs/microsoft-graph-or-azure-ad-graph/) Azure Active Directory (AAD) graph API. This change was necessary as a part of the transition, since it required behavior (requesting an additional access token) that's not exposed by cap/oidc.

## Testing

All existing tests are passing against the changes in this PR.

I've manually tested OIDC authentication using the following identity providers:
- Auth0
- Google
    - Tested with and without additional G Suite groups handling
- Azure Active Directory
    - Tested with and without additional 200+ groups handling

I also tested that the `namespace_in_state` parameter works as expected.

## Additional Context

The cap/oidc library introduces a [Request](https://github.com/hashicorp/cap/blob/main/oidc/request.go#L22) type that represents a single OIDC auth flow for a user. A `Request` is uniquely identified by a [State](https://github.com/hashicorp/cap/blob/main/oidc/request.go#L26). This PR includes changes to align terminology used in the plugin to that used in the cap/oidc library by renaming `State` to `Request` in certain instances. I'm happy to consider alternatives or leave the naming unchanged in this plugin.

## Related Pull Requests

- https://github.com/hashicorp/vault-plugin-auth-jwt/pull/155


path_oidc.go
```diff
@@ -404,119 +393,94 @@ func (b *jwtAuthBackend) authURL(ctx context.Context, req *logical.Request, d *f
...
-	if config.NamespaceInState && len(namespace) > 0 {
-		// embed namespace in state in the auth_url
-		stateID = fmt.Sprintf("%s,ns=%s", stateID, namespace)
-	}
 
-	authCodeOpts := []oauth2.AuthCodeOption{
-		oidc.Nonce(nonce),
+	// embed namespace in state in the auth_url
+	if config.NamespaceInState && len(namespace) > 0 {
+		stateWithNamespace := fmt.Sprintf("%s,ns=%s", oidcRequest.State(), namespace)
```
calvn, line 424: Don't we have to update the state in `oidcRequest` with this value?

path_oidc.go
```diff
@@ -136,13 +138,13 @@ func (b *jwtAuthBackend) pathCallbackPost(ctx context.Context, req *logical.Requ
 		},
 	}
 
-	// Store the provided code and/or token into state, which must already exist.
-	state, err := b.amendState(stateID, code, idToken)
+	// Store the provided code and/or token into its OIDC request, which must already exist.
+	oidcRequest, err := b.amendOIDCRequest(stateID, code, idToken)
```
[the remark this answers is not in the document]
austingebauer: I've changed vars named `oidcRequest` to `oidcReq` to avoid this here and elsewhere in 8efc627.

path_oidc.go
```diff
@@ -254,30 +239,34 @@ func (b *jwtAuthBackend) pathCallback(ctx context.Context, req *logical.Request,
...
+	if err := rawToken.Claims(&allClaims); err != nil {
+		return nil, err
 	}
+	delete(allClaims, "nonce")
 
-	if allClaims["nonce"] != state.nonce {
-		return logical.ErrorResponse(errTokenVerification + " Invalid ID token nonce."), nil
+	// The sub claim in the user info response must be verified to
+	// exactly match the sub claim in the ID token as described in:
+	// https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse
```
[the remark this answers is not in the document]
austingebauer: The comment is accurate, but it's really a validation detail of `provider.UserInfo()`. I've removed that detail and updated the comment to be more appropriate in 9b36341.

path_oidc.go
```diff
@@ -404,119 +393,94 @@ func (b *jwtAuthBackend) authURL(ctx context.Context, req *logical.Request, d *f
...
+	//
 	// TODO: it would be better to convey this to the UI and have it send the
 	// correct URL directly.
 	if config.OIDCResponseMode == responseModeFormPost {
 		redirectURI = strings.Replace(redirectURI, "ui/vault", "v1", 1)
 	}
 
-	provider, err := b.getProvider(config)
+	// Create a new OIDC request state
+	oidcRequest, err := b.createOIDCRequest(config, role, roleName, redirectURI, clientNonce)
```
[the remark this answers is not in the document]
austingebauer: I think that's a good idea. I've changed the sequence in 325bacd.

DrDaveD:
When this is getting pretty close to being final can you please say so here?  I would like to test it in my environment.

DrDaveD:
I also wonder if it's going to complicate or simplify adding code flow callback direct to vault #130 and device flow #131.

path_oidc.go
austingebauer: I originally thought the same, but the current behavior in the plugin doesn't update the actual state ID with the `,ns=<namespace>` suffix. You can see that the state ID is first created and set on [path_oidc.go#L483](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_oidc.go#L483), which occurs before the suffix is added on [path_oidc.go#L439](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_oidc.go#L439). It appears that the UI parses out the namespace suffix on [auth-jwt.js#L134](https://github.com/hashicorp/vault/blob/master/ui/app/components/auth-jwt.js#L134) to provide the proper state ID to the plugin callback handler.

provider_azure.go
```diff
@@ -3,22 +3,24 @@ package jwtauth
...
 	log "github.com/hashicorp/go-hclog"
 	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/clientcredentials"
 )
 
 const (
-	// The old MS graph API requires setting an api-version query parameter
-	windowsGraphHost  = "graph.windows.net"
-	windowsAPIVersion = "1.6"
+	// The host of the deprecated Azure Active Directory (AAD) graph API
```
calvn, line 18:
```suggestion
	// Deprecated: The host of the Azure Active Directory (AAD) graph API
```

path_oidc_test.go
```diff
@@ -838,7 +838,8 @@ func TestOIDC_Callback(t *testing.T) {
 			t.Fatal("nil response")
 		}
 
-		if !resp.IsError() || !strings.Contains(resp.Error().Error(), `error validating signature: oidc: expected audience "abc"`) {
+		expectedMsg := "invalid id_token: one audience (not_gonna_match) which is not the client_id (abc)"
+		if !resp.IsError() || !strings.Contains(resp.Error().Error(), expectedMsg) {
```
calvn, line 842: It looks like the library returns a specific typed error for this case, `ErrInvalidAuthorizedParty`, which you can use to check against instead.

calvn (approved):

path_oidc.go
calvn: Yeah, re-visited the logic and it makes sense now!

tvoran (approved):
I tried this with an azure setup and a user with 200 groups, and it still works 👍

path_oidc_test.go
austingebauer: Hmm, I gave this a try using `errors.Is(resp.Error(), oidc.ErrInvalidAuthorizedParty)`. It returned false. Looks like it's not wrapped because of how the error is included as a string to `logical.ErrorResponse()` on [path_oidc.go#L224](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/cap-oidc-lib/path_oidc.go#L224).

austingebauer:
@DrDaveD - You can go ahead and test this in your environment now. Please let me know if you find any regressions. I'm having a look at the PRs you've linked to see how this change may affect them. Will respond when I've gathered more context.

path_oidc_test.go
calvn: That's unfortunate... We could string check against `oidc.ErrInvalidAuthorizedParty.Error()` if we wanted the test to be a bit more robust, but this is fine too.

austingebauer: I like that better. Changed to check as you've suggested in d42d8d8.

austingebauer:
@DrDaveD - I took a look at the PRs that you've linked to. I don't see the usage of [cap/oidc](https://github.com/hashicorp/cap/tree/main/oidc) in this plugin as something that will complicate (or simplify..) the changes in the PRs. I do know that those use cases are being considered as a part of continued development on cap/oidc, so I expect that supporting them will eventually be simplified.

PR #159, opened 2021-02-11
Adds max_age role parameter and auth_time claim validation

austingebauer:
## Overview

This PR adds a new `max_age` parameter to roles. The `max_age` specifies the allowable elapsed time in seconds since the last time the user actively authenticated with the configured identity provider. 

When `max_age` is used, the ID Token returned must include an `auth_time` claim value (see [openid-connect-core-1_0.html#AuthRequest](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)). Validation of the `auth_time` claim value is provided by the cap/oidc library (see [oidc/provider.go#L414](https://github.com/hashicorp/cap/blob/main/oidc/provider.go#L414)).

The `max_age` for a role will be included in the generated auth URL as a query parameter, similar to the following example:

```
https://example.auth0.com/authorize?client_id=abc&max_age=60&nonce=n_NEJOgEMl4MBXf0LX6ES4&redirect_uri=https%3A%2F%2Fexample.com&response_type=code&scope=openid&state=st_o1myOrU9vQi6uiiR6SpU
```

## Testing

I've added a test for setting the `max_age` role parameter and generating an auth URL with the `max_age` query parameter.

I've also manually tested Azure Active Directory authentication using `max_age` to observe that active authentication was forced depending on the `max_age` and time of last active authentication.

### DIFF
--- a/path_oidc_test.go
+++ b/path_oidc_test.go
@@ -385,56 +385,46 @@ func TestOIDC_AuthURL_max_age(t *testing.T) {
 	require.NoError(t, err)
 	require.False(t, resp.IsError())
 
-	tests := []struct {
-		name           string
+	tests := map[string]struct {
 		maxAge         string
 		expectedMaxAge string
 		expectErr      bool
 	}{
-		{
-			name:           "auth URL for role with integer max_age of 60",
+		"auth URL for role with integer max_age of 60": {
 			maxAge:         "60",
 			expectedMaxAge: "60",
 		},
-		{
-			name:           "auth URL for role with integer max_age of 180",
+		"auth URL for role with integer max_age of 180": {
 			maxAge:         "180",
 			expectedMaxAge: "180",
 		},
-		{
-			name:           "auth URL for role with empty max_age",
+		"auth URL for role with empty max_age": {
 			maxAge:         "",
 			expectedMaxAge: "",
 		},
-		{
-			name:           "auth URL for role with duration string max_age of 30s",
+		"auth URL for role with duration string max_age of 30s": {
 			maxAge:         "30s",
 			expectedMaxAge: "30",
 		},
-		{
-			name:           "auth URL for role with duration string max_age of 2m",
+		"auth URL for role with duration string max_age of 2m": {
 			maxAge:         "2m",
 			expectedMaxAge: "120",
 		},
-		{
-			name:           "auth URL for role with duration string max_age of 1hr",
+		"auth URL for role with duration string max_age of 1hr": {
 			maxAge:         "1h",
 			expectedMaxAge: "3600",
 		},
-		{
-			name:      "auth URL for role with invalid duration string",
+		"auth URL for role with invalid duration string": {
 			maxAge:    "1hr",
 			expectErr: true,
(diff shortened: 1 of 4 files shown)


path_role.go
```diff
@@ -440,6 +447,10 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		role.VerboseOIDCLogging = verboseOIDCLoggingRaw.(bool)
 	}
 
+	if maxAgeRaw, ok := data.GetOk("max_age"); ok {
+		role.MaxAge = time.Duration(maxAgeRaw.(int)) * time.Second
```
pcman312, line 451: What about allowing for formatted strings such as `"1h"`? You could use [`parseutil.ParseDurationSecond`](https://github.com/hashicorp/vault/blob/master/sdk/helper/parseutil/parseutil.go#L17) for this.

pcman312, line 451: What are your thoughts on erroring/warning if `max_age` is a negative number?

path_oidc_test.go
```diff
@@ -365,6 +367,116 @@ func TestOIDC_AuthURL_namespace(t *testing.T) {
...
+			"oidc_client_id":     "abc",
+			"oidc_client_secret": "def",
+		},
+	}
+	resp, err := b.HandleRequest(context.Background(), req)
+	require.NoError(t, err)
+	require.False(t, resp.IsError())
+
+	tests := []struct {
+		name           string
```
pcman312, line 389: Since these test cases have a name, I'd prefer to see this collection as a map instead of a slice. It gives the name for free as the map key, and helps ensure there's no sharing of data between test runs since the tests are run in a different order each time.

path_role.go
austingebauer: Using duration strings is already supported by way of the field being a `TypeDurationSecond`. You can see that the tests in this PR provide some duration strings like `2m`, `1hr`. At this line, `maxAgeRaw` is the integer number of seconds that's already parsed from a given duration string (see [field_data.go#L213](https://github.com/hashicorp/vault/blob/master/sdk/framework/field_data.go#L213)).

austingebauer: The request validation for field data handles returning an error if `max_age` is a negative number (see [field_data.go#L219](https://github.com/hashicorp/vault/blob/master/sdk/framework/field_data.go#L219)). The error looks like: 
```
error converting input -2 for field "max_age": cannot provide negative value '-2'
```

path_oidc_test.go
austingebauer: Will do 👍

austingebauer: Done in 6928898.

path_role.go
pcman312: Oh interesting, I didn't realize the router did that conversion for us.

pcman312 (approved):

path_role.go
austingebauer: Yep, it's pretty convenient!

catsby (approved):
LGTM 👍

PR #182, opened 2021-10-14
Add a skip_browser argument to make auto-launching of the default browser optional

dekimsey:
# Overview
This change allows a user to opt-out of the auto-launching of their
default browser which may already have a logged in OIDC session. Since
SSO can be a bit tricky to logout/switch users, this provides the user
an escape option for more careful handling of the login attempt.

Primary use-case is Firefox Containers or users using separate browser
profiles to manage multiple logins. If one has an existing SSO session in their default browser,
it becomes impossible to get the vault CLI to login to another session.

# Design of Change
A command line flag was added to allow users to opt-out.

None

# Contributor Checklist
[x] https://github.com/hashicorp/vault/pull/12833
[x] Backwards compatible

**NOTE** I was unable to test the change, I could not find any docs on how to test the client side changes here. If I could get some instruction/direction on how to do that, I'd be happy to confirm it works as intended.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -301,7 +301,7 @@ Usage: vault login -method=oidc [CONFIG K=V...]
 Configuration:
 
   role=<string>
-      Vault role of type "OIDC" to use for authentication.
+    Vault role of type "OIDC" to use for authentication.
 
   listenaddress=<string>
     Optional address to bind the OIDC callback listener to (default: localhost).
@@ -316,10 +316,10 @@ Configuration:
     Optional callback host address to use in OIDC redirect_uri (default: localhost).
 
   callbackport=<string>
-      Optional port to to use in OIDC redirect_uri (default: the value set for port).
+    Optional port to to use in OIDC redirect_uri (default: the value set for port).
 
-  no-launch
-      Do not auto-launch default browser with OIDC login address.
+  skip_browser=<bool>
+    Toggle the automatic launching of the default browser to the login URL. (default: false).
 `
 
 	return strings.TrimSpace(help)


cli.go
```diff
@@ -74,6 +75,12 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 		callbackPort = port
 	}
 
+	doBrowserLaunch := defaultBrowserLaunch
+	_, ok = m["no-launch"]
```
austingebauer, line 79: I'm thinking we should check the value here, since we're documenting this option as a boolean. Otherwise, the no-launch behavior will still occur if you provide the default `no-launch=false` to the CLI.

cli.go
```diff
@@ -299,6 +311,9 @@ Configuration:
 
   callbackport=<string>
       Optional port to to use in OIDC redirect_uri (default: the value set for port).
+
+  no-launch
```
austingebauer, line 315: nit: I wonder if using `open-browser` (default of true) for the name would make the behavior of this option more clear? I see that our docs mention both opening and launching, so I'm okay with either. Curious to hear others' thoughts.

It would look like:
```
vault login -method=oidc -path=oidc role=test open-browser=false
```

cli.go
austingebauer: This `Auth` method is called from [login.go#L225](https://github.com/hashicorp/vault/blob/main/command/login.go#L225). See that there is an [interface](https://github.com/hashicorp/vault/blob/main/command/login.go#L13-L18) to implement which restricts this to using `map[string]string`. 

There is precedent for parsing string values as booleans using [strconv#ParseBool](https://pkg.go.dev/strconv#ParseBool), so I suggest we take that approach. Here is an example in Vault: [token/cli.go#L27](https://github.com/hashicorp/vault/blob/main/builtin/credential/token/cli.go#L27). Thanks 👍

dekimsey: Alright, made the update. Pending whatever changes get decided in the actual name.

Aside: I wanted to add a compile-time check for the interface, but that creates an import loop. On hindsight that's obvious. Is command.LoginHandler the only exported/pluggable interface in vault's cli?

cli.go
calvn: Since this is a new parameter, using the zero value of `false` as the default value might be a bit more intuitive. How about `skip_browser` with default of `false`?

cli.go
```diff
@@ -299,6 +317,9 @@ Configuration:
 
   callbackport=<string>
       Optional port to to use in OIDC redirect_uri (default: the value set for port).
+
```
austingebauer, line 320: Let's update the CLI help text name and description to use `skip_browser`.

cli.go
austingebauer: Thanks! LGTM. That's a good idea on the compile-time check, but you're right about the import loop. As far as I know, command.LoginHandler is the only interface for plugins to use for CLI handling.

calvn:
@austingebauer we should also update https://www.vaultproject.io/docs/auth/jwt#oidc-login-cli afterwards

austingebauer:
@calvn - The author has already provided a PR for the docs update: https://github.com/hashicorp/vault/pull/12833. Will review and get that merged too.

cli.go
dekimsey: Whoops! Updated help text.

I fixed whitespace on the other two options here since I am mucking around in the help anyway.

dekimsey:
Should I do the squash/rebase or is that done as part of the merge process?

cli.go
austingebauer: Ty!

austingebauer (approved):
@dekimsey - The squash is handled when merging, so you don't need to do it. I'll take care of merging this and getting it into a Vault release. Thanks again for the contribution!

PR #185, opened 2021-11-01
Revert a couple of Google deps back to what they were

kalafut:
We need additional time to validate Vault with them, since they're
used elsewhere.


austingebauer (approved):
👍

calvn (approved):

PR #192, opened 2021-12-20
Fixes UI for implicit flow by adding required source to postMessage

austingebauer:
## Overview

This PR fixes a regression introduced in Vault 1.9.1 that caused the implicit flow (`oidc_response_types=id_token`) to no longer work from the Vault UI. The cause of the regression was https://github.com/hashicorp/vault/pull/13133, which added a required `source` field to the data sent via `postMessage` between the window objects.

Fixes https://github.com/hashicorp/vault/issues/13460.

## Testing

I tested that this fixes the implicit flow using Azure AD from the Vault UI.

### DIFF
--- a/cli_responses.go
+++ b/cli_responses.go
@@ -449,7 +449,7 @@ func formpostHTML(path, code, state string) string {
       </div>
     </div>
 	<script>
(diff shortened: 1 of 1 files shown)


calvn (approved):

braunsonm:
Thanks for looking into this and the quick fix Austin!

PR #197, opened 2022-02-25
deps: downgrade google.golang.org/api to avoid forced upgrade on vault

calvn:
This PR reverts some of the deps on main to avoid forcing the vault repository to upgrade these dependencies. Similar in spirit to https://github.com/hashicorp/vault-plugin-auth-jwt/pull/185

```
$ go get google.golang.org/api@v0.29.0

$ go mod tidy
```


austingebauer (approved):
Ty!

PR #202, opened 2022-04-25
Return nil if no warnings when creating roles

AnPucel:
# Overview
A simple bug fix. The /role/ path doesn't return success because we get an empty secret interface returned instead of nil. To fix this we just return nil,nil when no warnings are present.

This affects those who are programmatically checking the cli commands to see if roles were successfully written. 

What is the change? 
return nil,nil when no warnings are present when creating roles

Why is the change needed?
So, users can know whether their roles were successfully written to vault

Users won't know if their roles were successfully written. If they want to programmatically check for it, they need to list all roles. 

# Design of Change
Return nil,nil if no warnings

[x] [Issue #15107](https://github.com/hashicorp/vault/issues/15107)

# Contributor Checklist
[My Docs PR Link](link)
[x] Backwards compatible

### DIFF
--- a/path_role.go
+++ b/path_role.go
@@ -561,6 +561,10 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 		return nil, err
 	}
 
+	if len(resp.Warnings) == 0 {
+		return nil, nil
+	}
+
 	return resp, nil
 }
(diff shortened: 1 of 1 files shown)


austingebauer (approved):
LGTM. Thanks, @AnPucel.

PR #204, opened 2022-05-12
Adds ability to use JSON pointer for the user claim

### LINKED ISSUE #180: [req] Allow json pointers when specifying a user claim
Hello,
Is it possible to allow for JSON pointers when specifying a `user_claim` in JWT roles? 
In some cases, the user is uniquely identified by a non root attribute, as for the following example:

```json
{
  "iat": 1629710930,
  "jti": "1a418f37-4358-4648-b41f-006e62ce0b87",
  "exp": 1629718130,
  "scope": "public",
  "user": {
     "email": "takiedd.mekhalfa@gmail.com",
     "full_name": "Taki Eddine MEKHALFA",
   }
}
```

It would be great if we could do `user_claime="/user/email"`.

I would be happy to work on this feature if you think it can be done without breaking something else.

Thank you,
Taki

austingebauer:
## Overview

This PR allows users to define the `user_claim` value as a JSON pointer. It introduces a new boolean parameter on the role, `UserClaimJSONPointer`, which allows users to opt in to the behavior. 

The decision to make this configurable was made to preserve existing behavior. For example, an existing role could have a `user_claim` value that is a valid JSON pointer to a different claim. This would result in an unexpected alias change. We want to prevent that from happening.

Closes #180

## Testing

I've added tests to this PR. I've also done some manual testing.


swenson (approved):
LGTM

fairclothjm (approved):

PR #208, opened 2022-07-18
VAULT-6727 Role resolution for JWT

VioletHynes:
Support role resolution functionality (to work with role based quotas). The intent of this operation is to take the same data as a login request and return the role that would be used in the request, so that we can quota upon login requests that have a certain role.

Afterwards, I will push a new tag and pull this new version into Vault.


path_login.go
```diff
@@ -35,40 +35,66 @@ func pathLogin(b *jwtAuthBackend) *framework.Path {
...
+		return "", nil, nil, err
 	}
 	if role == nil {
-		return logical.ErrorResponse("role %q could not be found", roleName), nil
+		return "", nil, logical.ErrorResponse("role %q could not be found", roleName), nil
 	}
 
 	if role.RoleType == "oidc" {
-		return logical.ErrorResponse("role with oidc role_type is not allowed"), nil
+		return "", nil, logical.ErrorResponse("role with oidc role_type is not allowed"), nil
```
austingebauer, line 81: I wonder if you also want to resolve roles for OIDC auth? The `pathLogin` handler only covers JWT auth. 

It's a bit confusing because OIDC and JWT auth are both implemented in this plugin and use the same role struct. The entry point for OIDC auth is at [path_oidc.go#L107](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/main/path_oidc.go#L107) in case you also want to resolve roles for it. Happy to help!

VioletHynes: Thanks for the comment, I did kind of miss that.

So, if I understand correctly, `login` gets you to JWT, and `oidc/auth_url` gets you to OIDC auth?

The quirk there is that all resolution does need to be done through the same request. From a quick look at the OIDC code, it might be as easy as just removing the highlighted error message from the resolve role path. Does that sound right to you?

VioletHynes: Just pushed a change that moves this logic into `pathLogin` explicitly, so role resolution should work for OIDC too - please re-review and let me know what you think. Thanks again for the call-out!

VioletHynes: It should be okay that the OICD auth flows don't make requests to `login`. The role resolution functionality is a step before the 'real request' is made, to determine what role is part of the request. The 'real request' to the actual login endpoint may not even happen, for example if the client is being rate limited. In other words, the two flows are distinct.

It shouldn't matter what the actual login API path is, as long as you can figure out the role from `login`.

See the following code snippet for how `core` figures out which role is being used: https://github.com/hashicorp/vault/blob/main/vault/core.go#L3351

austingebauer: Ah, that makes sense! Thanks for explaining :) Then yes, the change you made in 3f20995 should work.

austingebauer (approved):
LGTM!

PR #211, opened 2022-07-25
Azure provider - Add Support for US Gov L4 

MarkDordoy:
The azure provider has some code to correctly identify and convert the Azure AD Graph host to the MS Graph host. For reasons unknown to me, Microsoft will not update the JWT to correctly send MS Graph URLs even though AD Graph is deprecated.

This PR extends the current conversion to support the US Gov cloud offering (https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-request-differences)

### DIFF
--- a/provider_azure.go
+++ b/provider_azure.go
@@ -16,6 +16,7 @@ import (
 
 const (
 	// Deprecated: The host of the Azure Active Directory (AAD) graph API
+	// Conversion List: https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-request-differences 
 	azureADGraphHost   = "graph.windows.net"
 	azureADGraphUShost = "graph.microsoftazure.us"
 
--- a/provider_azure.go
+++ b/provider_azure.go
@@ -16,10 +16,12 @@ import (
 
 const (
 	// Deprecated: The host of the Azure Active Directory (AAD) graph API
-	azureADGraphHost = "graph.windows.net"
+	azureADGraphHost   = "graph.windows.net"
+	azureADGraphUShost = "https://graph.microsoftazure.us"
 
 	// The host and version of the Microsoft Graph API
 	microsoftGraphHost       = "graph.microsoft.com"
+	microsoftGraphUSHost     = "graph.microsoft.us"
 	microsoftGraphAPIVersion = "/v1.0"
 
 	// Distributed claim fields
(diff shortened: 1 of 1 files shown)


MarkDordoy:
@austingebauer tagging as i see you recently reviewed a previous PR. Sorry for direct tag.

provider_azure.go
```diff
@@ -16,10 +16,12 @@ import (
 
 const (
 	// Deprecated: The host of the Azure Active Directory (AAD) graph API
-	azureADGraphHost = "graph.windows.net"
+	azureADGraphHost   = "graph.windows.net"
+	azureADGraphUShost = "https://graph.microsoftazure.us"
```
austingebauer, line 20: Do you happen to know why the host would include the scheme (`https://`) in the case of `azureADGraphUShost` but not in `azureADGraphHost`?

MarkDordoy: Hello, sorry i did this by mistake in my haste to get a PR up. Should be updated now.

provider_azure.go
```diff
@@ -117,6 +119,9 @@ func (a *AzureProvider) getClaimSource(logger log.Logger, allClaims map[string]i
 	if urlParsed.Host == azureADGraphHost {
```
austingebauer, line 119: Can you add this link to the list here: https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-request-differences

MarkDordoy: Added

provider_azure.go
austingebauer: Are you able to confirm that you've tested this out and it works as expected? I don't have access to a US Gov L4 environment. This change seems simple enough, so I'm inclined to accept it.

provider_azure.go
austingebauer: Thank you!

provider_azure.go
austingebauer: @MarkDordoy - Just looking for a response here ^. I will merge this after 👍

MarkDordoy: Hey, I can confirm I tested out the us gov l4 environment response. It's as described in here. I even spoke to MS support about their plan to actually send the correct host but they couldn't provide when.

austingebauer: Excellent! Thanks for doing that. I was also wondering when they'd switch the distributed claims hosts over to MS graph.

austingebauer (approved):
LGTM. Thanks again, @MarkDordoy.

PR #226, opened 2023-03-26
Allow for role name to be used as entity alias

### LINKED ISSUE #217: [req] Ability to use JWT Role Name as Identity entity alias name
# Problem

There are certain situations where individual claims in the JWT tokens are insufficient at properly identifying a unique identity.

Take [GitLab JWTs](https://docs.gitlab.com/ee/ci/examples/authenticating-with-hashicorp-vault/#how-it-works) as an example. Each claim in the JWT is a distinct, disparate piece of information about which repo, branch, user, etc. caused the CI job to spawn.  

This makes it impossible to have granular Identities aliased to these JWTs. The finest grain Identity we can have in this example is `project_path` which severely limits the usefulness of the `bound_claims` restrictions that can be accomplished with roles. Building on this example, if we wanted to associate an identity with CI jobs in a certain repo but only on the `main` branch, that is impossible. 

This becomes increasingly difficult in Vault enterprise where the Identity engine is used to grant permissions within child namespaces. In order to solve this in the short term, we have mounted per-namespace auth backends and are having users create their own roles within their namespace-local backends which is potentially detrimental to our client count (core licensing metric).

# Proposed Solution

Allow for the name of the role used during authentication to be used as the Identity entity alias name.

This would allow us to leverage the powerful bound_claims restrictions and map them directly to an Identity.

For example, we could create a role like:
```json
{
  "name": "myteam.myproject.production"
  // add some boolean to toggle this behavior
  "role_name_as_entity_name": true
  "bound_claims": {
    "project_path": "myteam/myproject"
    "ref_protected": "true",
    "environment": "production",
    "environment_protected": "true"
  }
}
```
Then an Identity:
```json
{
  "name": "gitlab--myteam-myproject--prod"
}
```
And an alias:
```json
{
   "name":  "myteam.myproject.production",
   "canonical_id": "id-of-entity",
   "mount_accessor": "accessor-of-jwt-backend"
}
```

At this point, users are free to create internal groups (with the entity above as a member) in their namespaces with whatever policies they need. The auth mount can live in the root namespace, have access to any number of children, and we don't have to mount an auth backend per-namespace anymore.

mbillow:
# Overview

Allows for the name of the role being used during authentication to be provided as the name of the entity alias. This allows for JWTs with many distinct, disparate claims to effectively be mapped to a single identity. Examples and more explanation can be found in #217.

Closes #217. 

If this is something y'all would be willing to accept, I can make the documentation PR. 😄 

# Design of Change
Adds a boolean to roles on the JWT auth backend (`use_role_name_as_entity_alias`) which is checked at login and if true causes the name of the role being used to be provided as the entity alias name. This defaults to `false`.

- [ ] [Issue #217](https://github.com/hashicorp/vault-plugin-auth-jwt/issues/217)

# Testing Output
```
=== RUN   TestLogin_UseRoleNameAsEntityAlias
2023-03-26T14:00:29.546-0500 [WARN]  unable to locate /org/primary in claims: /org/primary at part 0: couldn't find key "org"
--- PASS: TestLogin_UseRoleNameAsEntityAlias (0.00s)
PASS
```

# Contributor Checklist
- [ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
- [x] Backwards compatible


verdel:
@fairclothjm, could you help review this PR? What needs to be done? Using the role name as an `entity-alias` will make the process of adding a Vault JWT role to an internal group more customizable. The role can essentially be added to the group based on `bound_claims` rather than `user_claim`. This approach allows for more precise matching by using multiple claims instead of just one.

This is described in more detail in the [issue](https://github.com/hashicorp/vault-plugin-auth-jwt/issues/217).

PR #245, opened 2023-07-19
chore: add errcheck for parsing cli flags

thevilledev:
# Overview

Add missing error check to `main()` for parsing CLI flags. If `flags.Parse()` fails now the error is hidden from the user.

# Design of Change

An error check was missing and I added it 👍

Originally noticed at https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/202

# Contributor Checklist
[-] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
Not needed

[-] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
Not needed

[x] Backwards compatible


austingebauer (approved):
Thanks!

PR #246, opened 2023-07-20
Add missing `Query: true` metadata to API definitions

maxb:
As described in hashicorp/vault#21949 endpoints with non-path parameters
that can be made use of during ReadOperation callbacks should have those
fields marked as `Query: true` to ensure the OpenAPI spec truly reflects
how the endpoints can be called.

This should also provide a proper fix to hashicorp/vault-client-go#155.

I'm also changing the operation ID to reflect that GET and POST on
`oidc/callback` are **not** at all the same operation with and without
parameters, but in fact play substantially different roles in the
overall OIDC login flow.


averche (approved):
LGTM, thanks again!

PR #255, opened 2023-10-12
Only use SIGTSTP on appropriate OSes

kpcraig:
In #251, we added `os.Kill` and `syscall.SIGTSTP` to the terminal interrupts we handled when terminating the listener. Unfortunately, `syscall.SIGTSTP` is not defined on all systems (notably, Windows).

The simplest way to fix this is to use [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) to separate the [SIGTSTP-having operating systems ](https://cs.opensource.google/search?q=SIGTSTP&ss=go%2Fgo:src%2Fsyscall%2F)  and the non-SIGTSTP systems. 

It also seems like there isn't an obvious alternative to SIGTSTP in Windows (if anyone knows...), so we just cut the behavior in the windows version.


maxcoulombe (approved):
Thanks you! Sorry for not catching that in the initial PR.

Pobably worth a changelog entry for a 1.17.2 release.

vinay-gopalan (approved):

raymonstah (approved):

PR #257, opened 2023-10-13
Add namespace parameter to callback url for cli vault login

### LINKED ISSUE #256: Vault CLI auth with oidc does not work with namespaces
Vault Version: `Vault 1.11.0+ent`

```
vault read -namespace=mynamespace auth/oidc/config 
Key                       Value
---                       -----
bound_issuer              n/a
default_role              access_readonly
jwks_ca_pem               n/a
jwks_url                  n/a
jwt_supported_algs        []
jwt_validation_pubkeys    []
namespace_in_state        true
oidc_client_id            XXXXXXXXXXXXXXXXXXXXXXXX
oidc_discovery_ca_pem     n/a
oidc_discovery_url        https://oidcprovider.example.com/api/v1/oauth
oidc_response_mode        n/a
oidc_response_types       []
provider_config           map[]
```

When I attempt to login using `-method=oidc -namespace=mynamespace` the namespace does not appear in the state parameter as I would expect.

[This cli code](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/eb63ad399383093b8e7f1bbdc54f84b8155b9e3c/cli.go#L229) makes the same request as the GUI but does not include the `?namespace=mynamespace` parameter

When I log in using the browser the `redirect_uri` contains the namespace and the `namespace_in_state` setting transfers it to the `state` parameter [here](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/eb63ad399383093b8e7f1bbdc54f84b8155b9e3c/path_oidc.go#L463).

jamie-pate:
Fixes #256

# Overview

Allows users that utilize namespaces to login with `vault login -namespace=mynamespace -method=oidc`

Currently the namespace is omitted from the redirect_uri when constructing the authURL for cli login. This change adds the namespace for feature parity with the GUI login.

# Design of Change

The current namespace is obtained from the client and added to `redirect_uri` parameter passed to the `/auth/{mount}/oidc/auth_url` request

[x] [Issue #256](https://github.com/hashicorp/vault-plugin-auth-jwt/256)

# Contributor Checklist
[x] Checked docs and this issue should 'just work'. Namespaces and `namespace_in_state` are [already mentioned](https://github.com/hashicorp/vault/blob/main/website/content/docs/auth/jwt/index.mdx#L75) in docs
[x] Backwards compatible

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"path"
@@ -115,7 +116,7 @@ func (h *CLIHandler) Auth(c *api.Client, m map[string]string) (*api.Secret, erro
 
 	role := m["role"]
 
-	authURL, clientNonce, err := fetchAuthURL(c, role, mount, callbackPort, callbackMethod, callbackHost)
+	authURL, clientNonce, err := fetchAuthURL(c, role, mount, callbackPort, callbackMethod, callbackHost, c.Namespace())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


jamie-pate:
After further testing I've found that this doesn't affect my ability to log in using vault CLI, the underlying problem was that my OP did not have the cli redirect_uri (`http://localhost:8250/oidc/callback`) configured, even though I'd configured it in the vault oidc config.

I'm still not clear why the namespace is passed in the state in the GUI but is not required in the CLI, but at least I'm able to log in with the CLI using oidc as intended.

PR #265, opened 2023-12-11
Support numeric claims

stefan-zh:
# Overview
The OIDC/JWT plugin does not support `bound_claims` other than strings or bools: https://github.com/hashicorp/vault-plugin-auth-jwt/blob/main/claims.go#L130. However, I encountered a JWT, which contains a numeric claim, which I want to bind to. 

I think it makes sense to support numeric bound claims - in fact, this was the original idea laid out in this comment: https://github.com/hashicorp/vault-plugin-auth-jwt/pull/24/files#r256926892

image

# Design of Change
We are adding support for numeric claims in `bound_claims`.

# Related Issues/Pull Requests:  
None

# Contributor Checklist
[X] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
No changes are needed in the documentation since it doesn't go into the details of what is/not supported in the bound claims.
[X] Backwards compatible
Numeric claims are currently not supported (they result in error), so this change only applies going forward.

### DIFF
--- a/claims.go
+++ b/claims.go
@@ -66,6 +66,7 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
 
 	switch v := val.(type) {
 	case float32:
+		return strconv.Itoa(int(v))
 	case float64:
 		return strconv.Itoa(int(v))
 	}
--- a/claims.go
+++ b/claims.go
@@ -4,6 +4,7 @@
 package jwtauth
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
@@ -54,21 +55,13 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
 	// The claims unmarshalled by go-oidc don't use UseNumber, so there will
 	// be mismatches if they're coming in as float64 since Vault's config will
 	// be represented as json.Number. If the operator can coerce claims data to
-	// be in string form, there is no problem. Alternatively, we could try to
-	// intelligently convert float64 to json.Number, e.g.:
-	//
-	// switch v := val.(type) {
-	// case float64:
-	// 	val = json.Number(strconv.Itoa(int(v)))
-	// }
-	//
-	// Or we fork and/or PR go-oidc.
-
+	// be in string form, there is no problem. As an alternative, we try to
+	// intelligently convert float32 and float64 to json.Number:
 	switch v := val.(type) {
 	case float32:
-		return strconv.Itoa(int(v))
+		return json.Number(strconv.Itoa(int(v)))
 	case float64:
-		return strconv.Itoa(int(v))
+		return json.Number(strconv.Itoa(int(v)))
 	}
 	return val
 }
@@ -129,12 +122,12 @@ func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims,
 
 		actVals, ok := normalizeList(actValue)
 		if !ok {
-			return fmt.Errorf("received claim is not a string or list: %v", actValue)
+			return fmt.Errorf("received claim is not a string, bool, int or list: %v", actValue)
 		}
 
 		expVals, ok := normalizeList(expValue)
 		if !ok {
-			return fmt.Errorf("bound claim is not a string or list: %v", expValue)
+			return fmt.Errorf("bound claim is not a string, bool, int or list: %v", expValue)
 		}
 
 		found, err := matchFound(expVals, actVals, useGlobs)
@@ -175,7 +168,7 @@ func matchFound(expVals, actVals []interface{}, useGlobs bool) (bool, error) {
 	return false, nil
 }
 
-// normalizeList takes a string, bool or list and returns a list. This is useful when
+// normalizeList takes a string, bool, json.Number or list and returns a list. This is useful when
 // providers are expected to return a list (typically of strings) but reduce it
 // to a string type when the list count is 1.
 func normalizeList(raw interface{}) ([]interface{}, bool) {
@@ -184,7 +177,7 @@ func normalizeList(raw interface{}) ([]interface{}, bool) {
 	switch v := raw.(type) {
 	case []interface{}:
 		normalized = v
-	case string, bool:
+	case string, bool, json.Number:
 		normalized = []interface{}{v}
 	default:
 		return nil, false
--- a/claims.go
+++ b/claims.go
@@ -6,6 +6,7 @@ package jwtauth
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


claims.go
```diff
@@ -63,6 +64,11 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
 	//
 	// Or we fork and/or PR go-oidc.
 
+	switch v := val.(type) {
+	case float32:
+	case float64:
```
austingebauer, line 69: This should be `case float32, float64:` on a single line to cover both types.

claims.go
```diff
@@ -63,6 +64,11 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
...
+		return strconv.Itoa(int(v))
```
austingebauer, line 70: Should we wrap this in `json.Number()` so that it matches with the type that comes in from Vault's API? Similar to what's in the comment above.

claims.go
```diff
@@ -63,6 +64,11 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
 	//
 	// Or we fork and/or PR go-oidc.
```
austingebauer, line 65: We should modify this comment to remove the suggested approach since we're going to use it. I do think it's worth leaving part of this comment in place to explain the non-obvious bits around json.Number.

claims_test.go
```diff
@@ -37,10 +37,10 @@ func TestGetClaim(t *testing.T) {
 		claim string
 		value interface{}
 	}{
-		{"a", float64(42)},
-		{"/a", float64(42)},
+		{"a", "42"},
```
austingebauer, line 40: Can we use `json.Number("42")` around these instead?

austingebauer (commented):
Left a few comments but it's mostly looking good. I think this will be a nice addition to bound claims. Would you also be open to creating a docs PR in Vault to mention that numeric claims are supported? If not, I can handle it.

claims.go
stefan-zh: Unfortunately, the combined case doesn't work because of limitation in the Go compiler for multiple types in a `switch case`: https://github.com/golang/go/issues/36616#issuecomment-575451270. 

I can use `fallthrough`? https://stackoverflow.com/a/40822747/9698467

claims.go
stefan-zh: I don't think we can - the rest of the code expects the claim to be `string`, `bool` or `[]`: https://github.com/hashicorp/vault-plugin-auth-jwt/blob/main/claims.go#L125 That is why I leave it `string`.

stefan-zh:
Thanks for the review @austingebauer ! I responded to the items in the comments, so let me know what you think and agree on how to proceed.

claims.go
austingebauer: Ah, missed that detail. Thanks for the context! What you've got now looks good to me.

claims.go
austingebauer: I tested this out using the Vault API with a numeric bound claims value and it doesn't work right now. The reason is that in `validateBoundClaims` the actual type becomes `string` and expected type is `json.Number`. This causes `matchFound` to return false.

This patch makes the two types match up as `json.Number`.

```
diff --git a/claims.go b/claims.go
index a032a29..d0fb028 100644
--- a/claims.go
+++ b/claims.go
@@ -4,6 +4,7 @@
 package jwtauth
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
@@ -66,9 +67,9 @@ func getClaim(logger log.Logger, allClaims map[string]interface{}, claim string)
 
 	switch v := val.(type) {
 	case float32:
-		return strconv.Itoa(int(v))
+		return json.Number(strconv.Itoa(int(v)))
 	case float64:
-		return strconv.Itoa(int(v))
+		return json.Number(strconv.Itoa(int(v)))
 	}
 	return val
 }
@@ -184,7 +185,7 @@ func normalizeList(raw interface{}) ([]interface{}, bool) {
 	switch v := raw.(type) {
 	case []interface{}:
 		normalized = v
-	case string, bool:
+	case string, bool, json.Number:
 		normalized = []interface{}{v}
 	default:
 		return nil, false
```

I'd be in support of moving forward with this if you agree.

claims.go
```diff
@@ -122,12 +130,12 @@ func validateBoundClaims(logger log.Logger, boundClaimsType string, boundClaims,
 
 		actVals, ok := normalizeList(actValue)
 		if !ok {
-			return fmt.Errorf("received claim is not a string or list: %v", actValue)
+			return fmt.Errorf("received claim is not a string, bool, int or list: %v", actValue)
```
stefan-zh, line 133: in this (and the next) error message I decided to use `int` instead of `json.Number` because this error prints to terminal and I think from user perspective `int` is more clear. Not sure what you think.

claims.go
stefan-zh: OK sounds good to me! I just pushed a new commit to use `json.Number` and add the new type in the switch case.

claims.go
austingebauer: Makes sense to me. The types being `json.Number` is an internal detail 👍

claims.go
austingebauer: Thanks!

austingebauer (approved):
Looks good after we add the bit of test coverage! Would you mind adding a line to the changelog for this enhancement? You can include it in this PR under "unreleased".

Thanks for this contribution, @stefan-zh 🎉

claims_test.go
```diff
@@ -37,10 +37,10 @@ func TestGetClaim(t *testing.T) {
 		claim string
```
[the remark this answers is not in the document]
stefan-zh: Thanks for the test cases. I added them!

stefan-zh:
I added the test cases and added a changelog entry under `Unreleased`. Thank you for your review and helpful suggestions @austingebauer !

stefan-zh:
Hi @austingebauer , I hope the things I added look good to you. I don't have permission to merge this code in, so could you do this for me? Thank you!

austingebauer:
@stefan-zh - Merging this now. Thanks again! 

Any chance you'd be able to update docs in the [Vault repo](https://github.com/hashicorp/vault/tree/main/website#vault-website) related to this? No worries if not 🙂

stefan-zh:
Hey @austingebauer thank you for merging it in! I will create a PR on the `vault` repo to update the docs. One place that should be updated is: https://developer.hashicorp.com/vault/api-docs/auth/jwt#bound_claims I will take a more detailed look tomorrow.

Issue #128, opened 2020-08-04
JWT Auth config discrepancy between the documentation and code


afrosimon:
I'm running a local Keycloak instance which provides me a JWT and I've run into a confusing situation when configuring vault to accept these JWTs.

As per the official documentation (https://www.vaultproject.io/docs/auth/jwt#jwt-authentication) I've ran this command:

```
vault write auth/jwt/config \
    oidc_discovery_url="https://myco.auth0.com/" \
    oidc_client_id="m5i8bj3iofytj" \
    oidc_client_secret="f4ubv72nfiu23hnsj" \
    default_role="demo"
```

And then I created the role (no issues there AFAIK). And when I tried to login, I hit this error:

```
vault write auth/jwt/login role=demo jwt=SOME_JWT
Error writing data to auth/jwt/login: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/jwt/login
Code: 500. Errors:

* unhandled case during login
```

I've eventually found where this error is raised (https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_login.go#L213). So the switch case is based on authType and based on my reading of that function, providing the OIDC URL + client ID + client secret puts the authType as OIDCFlow - which is not caught by the aforementioned switch case and that's why I'm getting the error! If I do **not** provide the oidc_client_id & oidc_client_secret the authType is OIDCDiscovery and I can log in successfully.

So my question is, is it simply an issue with the vault documentation? Feels like I'm missing something here..

dev-e:
https://github.com/hashicorp/vault-plugin-auth-jwt/issues/72#issuecomment-526746497

Eyal-Shalev:
@afrosimon THANK YOU VERY VERY MUCH.
Hours of headaches solved by your investigation.

austingebauer:
This has been addressed in both [documentation](https://developer.hashicorp.com/vault/docs/auth/jwt#configuration) and [code](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/main/backend.go#L154). I'm going to close this, but feel free to reopen if you still think there is an issue.

PR #277, opened 2024-02-03
Allow for multiple keysets configured

johnlanda:
# Overview

This change allows for the JWT auth backend to be configured with multiple JWKS URLs as the result of a customer request.

This change should allow for current configuration using `jwks_url` and `jwks_ca_pem` to work as normal. Users may optionally choose to instead specify sets of `jwks_pairs`. If `jwks_pairs` are specified then the underlying `Validator` from the cap library will loop over the available `jwks_pairs` in order to attempt to validate the JWT.

# Design of Change
This change updates the config for the `jwtAuthBackend` in the `path_config.go` file. It also updates the `backend.go` file to ensure that when using `jwks_pairs` we iterate over the JWKS uris and CA PEMs to initialize all key sets.

[ ] [Cap PR #128](https://github.com/hashicorp/cap/pull/128)

# Contributor Checklist
[My Docs PR Link](link)
[x] Backwards compatible


backend.go
```diff
@@ -145,15 +146,28 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
...
+	var keySets []jwt.KeySet
 
 	// Configure the key set for the validator
 	switch config.authType() {
 	case JWKS:
 		keySet, err = jwt.NewJSONWebKeySet(b.providerCtx, config.JWKSURL, config.JWKSCAPEM)
+		keySets = []jwt.KeySet{keySet}
+	case MultiJWKS:
+		for k, v := range config.JWKSPairs {
+			// TODO (@johnlanda): read this in as map[string]string in config so we don't need type conversion.
```
johnlanda, line 158: Is there a pattern for nested config in the path config? Ideally I would like these to be more like a list of structs than a map of key-value pairs.

go.mod
```diff
@@ -97,3 +97,5 @@ require (
 	google.golang.org/protobuf v1.31.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hashicorp/cap v0.4.1 => /Users/john.landa/Dev/cap
```
johnlanda, line 101: To be replaced once the Cap PR is merged.

https://github.com/hashicorp/cap/pull/128

path_config.go
```diff
@@ -72,6 +73,10 @@ func pathConfig(b *jwtAuthBackend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: "The CA certificate or chain of certificates, in PEM format, to use to validate connections to the JWKS URL. If not set, system certificates are used.",
 			},
+			"jwks_pairs": {
```
johnlanda, line 76: Like the comment above: it would probably be a more usable config if there is a good way to do a slice of struct here instead of a map. I was having a hard time getting that to parse properly.

path_config.go
```diff
@@ -255,10 +262,13 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 	if config.JWKSURL != "" {
 		methodCount++
 	}
+	if len(config.JWKSPairs) > 0 {
```
johnlanda, line 265: I left the `jwks_url` and `jwks_ca_pem` config in place to maintain backwards compatibility. It could be nice just to specify a single `jwks_pair` instead but I think the backwards compatibility was more important.

path_config.go
```diff
@@ -279,30 +289,27 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
...
+			if r := b.validateJWKSURL(ctx, k, v.(string)); r != nil {
+				return r, nil
+			}
 		}
 
 	case len(config.JWTValidationPubKeys) != 0:
 		for _, v := range config.JWTValidationPubKeys {
 			if _, err := certutil.ParsePublicKeyPEM([]byte(v)); err != nil {
-				return logical.ErrorResponse(errwrap.Wrapf("error parsing public key: {{err}}", err).Error()), nil
+				return logical.ErrorResponse(fmt.Errorf("error parsing public key: %w", err).Error()), nil
```
johnlanda, line 312: updating deprecated method

path_config.go
```diff
@@ -403,6 +410,29 @@ func (b *jwtAuthBackend) createCAContext(ctx context.Context, caPEM string) (con
 	return caCtx, nil
 }
 
+func (b *jwtAuthBackend) validateJWKSURL(ctx context.Context, JWKSURL, JWKSCAPEM string) *logical.Response {
```
johnlanda, line 413: This is just the existing logic split into a helper function to be used for the single and multi JWKS URL cases.

backend.go
austingebauer: Unfortunately there isn't a great pattern for complex/nested data types in the Vault API. You can use `framework.TypeMap` as you have, but it has the limitation that it's not usable by the Vault CLI without converting the entire request body to JSON. Happy to show you an example of this limitation.

I'm wondering if using `framework.TypeKVPairs` would be better. That would make the data type `map[string]string` where the key is the JWKS URL and the value the CA PEM. It also allows for using the Vault CLI more naturally.

path_config.go
austingebauer: Ya.. it's a challenge today with Vault backends. There are some ways to make it work but they require us to manually parse outside of the backend framework. Do we anticipate needing any configuration in addition to the JWKS URL and CA?

path_config.go
austingebauer: +1 on that decision

path_config.go
johnlanda: I don't expect any additional config to be added to the pairs.

backend.go
johnlanda: `TypeKVPairs` sounds like an improvement over the current setup at least. I don't think the config files will read super intuitively unfortunately since the fields within each `jwks_pair` will be effectively unlabeled.

austingebauer: Ya, I see what you're saying. Do you have a rough example of what that config file would look like? It might help me think of something better 🤔

johnlanda: If there is anyway to make it like a repeated block in yaml where it would be something like 
```
jwks_pairs:
  - jwks_url: www.foobar.com/wellknown
    jwks_ca_pem: adfjkafklasdfs232flfdskja
  - jwks_url: www.barbaz.com/wellknown
    jwks_ca_pem: aweeqrrnrqewrkljerwqwr
```

johnlanda: Looking through the nomad docs for setting up vault doesn't show setting the ca cert. Is that something that vault does automatically? If so we could just make this a list of urls

austingebauer: I'm trying something out with the repeated block idea. I'll follow up here shortly!

> Looking through the nomad docs for setting up vault doesn't show setting the ca cert. Is that something that vault does automatically? If so we could just make this a list of urls

Able to share a link to which doc you're looking at? It's not required if your IdP (Nomad in this case) is not TLS protected or the system trust store where Vault is running has already trusted the CA of the IdP.

jkirschner-hashicorp: > Looking through the nomad docs for setting up vault doesn't show setting the ca cert. Is that something that vault does automatically?

Even if Nomad doesn't now, or maybe doesn't in its "easy getting starts commands / docs", setting the CA cert might be something we want to support (even if some people leave it as an empty string). I defer to you both, though.

johnlanda: Updated so that the config for `jwks_pairs` is effectively a slice of structs and the fields are named for easier reading of config.

johnlanda: New config looks like this
```vault write auth/jwt/config -<<EOF 
{
    "jwks_pairs": [
        { "jwks_url": "foobar.com", "jwks_ca_pem": "asdfadsfasdf" },
        { "jwks_url": "barbaz.com", "jwks_ca_pem": "adfsasdfasdf" }
    ]
}
EOF
```

austingebauer (commented):
Leaving a couple of small comments as I'm looking through this. Looking good!

jwks_pairs.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/mitchellh/mapstructure"
+)
+
+type JWKSPair struct {
+	JWKSUrl   string `mapstructure:"jwks_url"`
+	JWKSCAPEM string `mapstructure:"jwks_ca_pem"`
+}
+
+func NewJWKSPairsConfig(jc *jwtConfig) ([]*JWKSPair, error) {
+	if len(jc.JWKSPairs) <= 0 {
```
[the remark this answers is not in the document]
johnlanda: I often use `<=` instead of `==` as it is more restrictive than `==`. Though `len` should always return a positive integer so I'll defer to your preference.

jwks_pairs.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil, nil
+	}
+
+	pairs := make([]*JWKSPair, 0, len(jc.JWKSPairs))
+	for i := 0; i < len(jc.JWKSPairs); i++ {
+		jp, err := Initialize(jc.JWKSPairs[i].(map[string]interface{}))
```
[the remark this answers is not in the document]
johnlanda: Ack. Will update and add a test.

backend.go
austingebauer: Thanks for the example, @johnlanda!

jwks_pairs.go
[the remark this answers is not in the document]
johnlanda: Added in f7733e98b32797fde5277d2056f794e8d53da445

jwks_pairs.go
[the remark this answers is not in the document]
johnlanda: This is probably an overly defensive habit so as not to assume that the function called always returns a positive number.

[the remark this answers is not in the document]
austingebauer: Makes sense! I think it's okay to leave as is :)

backend.go
```diff
@@ -145,15 +146,29 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
...
+		keySets = []jwt.KeySet{keySet}
+	case MultiJWKS:
+		pairs, err := NewJWKSPairsConfig(config)
+
+		for _, p := range pairs {
+			keySet, keySetErr := jwt.NewJSONWebKeySet(b.providerCtx, p.JWKSUrl, p.JWKSCAPEM)
+			if keySetErr != nil {
+				err = multierror.Append(err, keySetErr)
+			}
+			keySets = append(keySets, keySet)
```
austingebauer, line 164: I believe the error is shadowed here such that if errors occurred, they wouldn't be appended to the outer error. I don't think we'd end up in the `if err != nil {` check below this switch case.

backend.go
```diff
@@ -145,15 +146,29 @@ func (b *jwtAuthBackend) jwtValidator(config *jwtConfig) (*jwt.Validator, error)
...
 	case JWKS:
 		keySet, err = jwt.NewJSONWebKeySet(b.providerCtx, config.JWKSURL, config.JWKSCAPEM)
+		keySets = []jwt.KeySet{keySet}
+	case MultiJWKS:
+		pairs, err := NewJWKSPairsConfig(config)
+
+		for _, p := range pairs {
+			keySet, keySetErr := jwt.NewJSONWebKeySet(b.providerCtx, p.JWKSUrl, p.JWKSCAPEM)
+			if keySetErr != nil {
+				err = multierror.Append(err, keySetErr)
```
austingebauer, line 162: Missing a `continue`? A nil `keySet` would be appended for the error case.

backend.go
johnlanda: Good catch. Looks like we can rename that error to something like `pairsErr` and just check and return it before the loop if not nil.

backend.go
johnlanda: 🤦

austingebauer (approved):
Thanks again for this contribution, @johnlanda! LGTM

PR #290, opened 2024-04-04
Update testJWT to contain a valid signature

jasonodonnell:
Go Jose added additional scrutiny to token signatures in v4 and HS256 is no longer a valid algorithm. Our hardcoded test JWT uses HS256 so I've swapped it out with RS256.

Verified the tests failed before the signature and pass after.


go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/hashicorp/vault-plugin-auth-jwt
 
-go 1.20
+go 1.21
```
tvoran, line 3: Don't forget to bump .go-version as well.

jasonodonnell: Good call!

.go-version
```diff
@@ -1 +1 @@
-1.20.4
+1.21.1
```
tvoran, line 1:
```suggestion
1.21.9
```

tvoran (approved):

PR #313, opened 2024-08-05
Add the ability to ignore unhandled critical certificate exceptions

sgmiller:
Some customers have complex PKI setups that include certificate extensions
marked "critical" that the Go TLS stack doesn't understand.  When it
encounters such an extension, it fails TLS validation. 

This PR adds an option to the JWT auth method,
`unsupported_critical_cert_extensions`, which can be set to a comma
separated list of the OIDs of extensions that should be ignored.


path_config.go
```diff
@@ -374,9 +387,40 @@ func (b *jwtAuthBackend) createProvider(config *jwtConfig) (*oidc.Provider, erro
 		supportedSigAlgs = []oidc.Alg{oidc.RS256}
 	}
 
+	var opts []oidc.Option
+	if len(config.UnsupportedCriticalCertExtensions) > 0 {
```
tomcf-hcp, line 391: If `len(config.UnsupportedCriticalCertExtensions) == 0` and `config.OIDCDiscoveryCAPEM == ""`, then this is all a no-op, and opts is then empty, correct?
Is that the same behavior as before, since you've replaced `oidc.WithProviderCA(config.OIDCDiscoveryCAPEM)` with `opts...`?

sgmiller: Yeah, it was weird to me that option was being provided when the config entry was empty.  The underlying field in the cap library is a string, so it's zero value is "", so it's okay to have provided it with the option setting it to "" but not super clean.

fairclothjm (approved):
LGTM

PR #323, opened 2025-03-25
Optionally Ignore trailing "/" at end of boundAudience

jaireddjawed:
# Overview
This PR allows users to optionally ignore the trailing slash at the end of each bound audience (if the trailing slash exists) by setting the `bound_audience_disregard_trailing_slash` flag to `true`. The PR also prevents duplicate bound_audiences from being entered, where the only difference would be the trailing slash (ie adding `https://testing.com` and `https://testing.com/` to the `bound_audiences` field results in `https://testing.com` being the only one added when the flag is in place. The default behavior remains in place if the flag is not set to true.

# Testing

* Created API on Auth0 with audience of `https://testing.com`

### Use without trailing slash and without flag (shows that default behavior works normally)

* Enable JWT in Vault

```vault auth enable jwt```

* Config JWT

```
vault write auth/jwt/config \
    oidc_discovery_url="https://MY_DOMAIN.us.auth0.com/" \
    oidc_client_id="" \
    oidc_client_secret="" \
    default_role="demo"
```

* Create Role on Vault

```
vault write auth/jwt/role/my-role \
    role_type="jwt" \
    user_claim="sub" \
    bound_audiences='https://testing.com' \
     policies="default"
```

* Read role from vault

```
vault read auth/jwt/role/my-role
Key                        Value
---                        -----
allowed_redirect_uris      <nil>
bound_audiences            [https://testing.com]
bound_claims               <nil>
bound_claims_type          string
bound_subject              n/a
claim_mappings             <nil>
clock_skew_leeway          0
expiration_leeway          0
groups_claim               n/a
max_age                    0
not_before_leeway          0
oidc_scopes                <nil>
policies                   [default]
role_type                  jwt
token_bound_cidrs          []
token_explicit_max_ttl     0s
token_max_ttl              0s
token_no_default_policy    false
token_num_uses             0
token_period               0s
token_policies             [default]
token_ttl                  0s
token_type                 default
user_claim                 sub
user_claim_json_pointer    false
verbose_oidc_logging       false
```

* Retrieve JWT Token from Auth0

```
curl --request POST \
  --url 'https://MY_DOMAIN.us.auth0.com/oauth/token' \
  --header 'content-type: application/json' \
  --data '{
    "client_id": "AUTH0_CLIENT_ID",
    "client_secret": "AUTH0_CLIENT_SECRET",
    "audience": "https://testing.com",
    "grant_type": "client_credentials"
  }'
```

* Successful authentication into Vault with the given jwt token from Auth0

```
[...]

PR #324, opened 2025-04-01
Vault 33101 backport to 1.16.x

jaireddjawed:
Backport the optional ignoring of a trailing "/" in any bound_audience. [Original PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/323/).


path_role.go
```diff
@@ -120,6 +120,10 @@ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 				Type:        framework.TypeCommaStringSlice,
 				Description: `Comma-separated list of 'aud' claims that are valid for login; any match is sufficient`,
 			},
+			"normalize_bound_audiences": {
```
tvoran, line 123: Why is this "normalize_bound_audiences" here, but "bound_audience_disregard_trailing_slash" [in main](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/e9b60da70def1222ba843987c6c128567ee33747/path_role.go#L123)?

tvoran (commented):
Just in case you aren't aware, typically when backporting we'd want to start by cherry-pick'ing the commit that went into main, and only make changes if there's a merge conflict.

So in this case you'd cherry-pick the commit that went into main (e9b60da):
```
git checkout release/vault-1.16.x
git pull
git checkout -b VAULT-33101-backport-1.16.x
git cherry-pick e9b60da
```

It's also best to keep dependency and go updates separate from code changes, to make reverting changes easier.

path_role.go
jaireddjawed: That name was a leftover from an old version. @benashz wanted me to change it to `normalize_bound_audiences` since `bound_audience_disregard_trailing_slash` is too large of a flag name. But by then, the original PR had already been approved and merged into main. He said that changing it should be fine since there hasn't been any official release.

jaireddjawed:
> Just in case you aren't aware, typically when backporting we'd want to start by cherry-pick'ing the commit that went into main, and only make changes if there's a merge conflict.
> 
> So in this case you'd cherry-pick the commit that went into main from the [original PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/323) ([e9b60da](https://github.com/hashicorp/vault-plugin-auth-jwt/commit/e9b60da70def1222ba843987c6c128567ee33747)):
> 
> ```
> git checkout release/vault-1.16.x
> git pull
> git checkout -b VAULT-33101-backport-1.16.x
> git cherry-pick e9b60da
> ```
> 
> It's also best to keep dependency and go updates separate from code changes, to make reverting changes easier.

Understood. I'll cherry-pick the commits from now on.

path_role.go
jaireddjawed: Here is the [PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/331) that changes the flag's name in main.

CHANGELOG.md
```diff
@@ -38,6 +38,23 @@ IMPROVEMENTS:
   * `golang.org/x/sync` v0.5.0 -> v0.6.0
   * `google.golang.org/api` v0.154.0 -> v0.163.0
 
+## v0.19.1
+
+FEATURES:
+  * Added `normalize_bound_audiences` as a flag that can be enabled to ignore the trailing slash at the end of each bound audience. Bound audiences that differ by only a trailing slash are treated as equivalent. This feature is set to `false` by default.
```
digivava, line 44: I would call this an Improvement not a Feature (see my comment on the main changelog PR)

PR #329, opened 2025-04-01
allow optional access token to be used to fetch additional groups

hashiblaum:
this allows an optional access token , `ms_graph_access_token`, to be specified while logging in using the JWT auth backend. This token will only be used with the azure provider and is used to fetch additional groups from the Microsoft Graph API

Testing:

was able to successfully login and retrieve group memberships as indicated by the policies attached to the token. these policies were attached to group-alias's corresponding to the Microsoft group IDs that were pulled with the access token

```
vault write auth/jwt/login jwt=$ID_TOKEN ms_graph_access_token=$ACCESS_TOKEN
Key                  Value
---                  -----
token                {}
token_accessor       {}
token_duration       768h
token_renewable      true
token_policies       ["default"]
identity_policies    ["test-policy1" "test-policy2"]
policies             ["default" "test-policy1" "test-policy2"]
token_meta_role      test-entra-jwt
```

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -34,6 +34,10 @@ func pathLogin(b *jwtAuthBackend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: "The signed JWT to validate.",
 			},
+			"ms_graph_access_token": {
+				Type:        framework.TypeString,
+				Description: "An optional token used to fetch group memberships specified by the claim source in the jwt",
+			},
 		},
 
 		Operations: map[logical.Operation]framework.OperationHandler{
@@ -112,6 +116,8 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 		return logical.ErrorResponse("missing token"), nil
 	}
 
+	msGraphAccessToken := d.Get("ms_graph_access_token").(string)
+
 	if len(role.TokenBoundCIDRs) > 0 {
 		if req.Connection == nil {
 			b.Logger().Warn("token bound CIDRs found but no connection information available for validation")
@@ -172,8 +178,7 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d 
 			}
 		}
 	}
-
-	alias, groupAliases, err := b.createIdentity(ctx, allClaims, roleName, role, nil)
+	alias, groupAliases, err := b.createIdentity(ctx, allClaims, roleName, role, &accessTokenSrc{accessToken: msGraphAccessToken})
 	if err != nil {
 		return logical.ErrorResponse(err.Error()), nil
 	}
@@ -350,3 +355,15 @@ const (
 Authenticates JWTs.
 `
 )
+
+type accessTokenSrc struct {
+	accessToken string
+}
+
+func (j *accessTokenSrc) Token() (*oauth2.Token, error) {
+	return &oauth2.Token{
+		AccessToken: j.accessToken,
+		TokenType:   "Bearer",
+	}, nil
+
+}


path_login.go
```diff
@@ -34,6 +34,10 @@ func pathLogin(b *jwtAuthBackend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: "The signed JWT to validate.",
 			},
+			"ms_graph_access_token": {
```
tvoran, line 37: Seems like this could potentially be used by other providers in the future; maybe something like "distributed_claim_access_token", "distributed_groups_access_token", etc. instead of specifically MS.

tvoran (commented):
I assume we've considered adding the access token as an option in the jwt config, like in the `provider_config` block (similar to [oidc config](https://developer.hashicorp.com/vault/docs/auth/jwt/oidc-providers/azuread#optional-azure-specific-configuration))?

path_login.go
hashiblaum: good call, will rename the variable here.

Also I didn't consider adding it to the config, since this access token has a TTL associated with it and it is normally generated at the same time the id_token is generated. so it makes sense that they are supplied together during jwt login.

path_login.go
```diff
@@ -112,6 +116,8 @@ func (b *jwtAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return logical.ErrorResponse("missing token"), nil
 	}
 
+	msGraphAccessToken := d.Get("distributed_claim_access_token").(string)
```
tvoran, line 119: We should probably change this variable name now too.

tvoran (commented):
Don't forget to update the description after changing the parameter name. 

Can you also include the jwt config and role config in your testing example? Have we tested this when there's no access token or distributed groups claim? Like [with k8s](https://developer.hashicorp.com/vault/docs/auth/jwt/oidc-providers/kubernetes).

hashiblaum:
> Don't forget to update the description after changing the parameter name.
> 
> Can you also include the jwt config and role config in your testing example? Have we tested this when there's no access token or distributed groups claim? Like [with k8s](https://developer.hashicorp.com/vault/docs/auth/jwt/oidc-providers/kubernetes)

Is the description `An optional token used to fetch group memberships specified by the claim source in the jwt` not valid? should I mention distributed claims somewhere in there?

Also updated testing with the configs and was able to use the jwt auth backend with tokens generated from k8s

tvoran:
> Is the description `An optional token used to fetch group memberships specified by the claim source in the jwt` not valid? should I mention distributed claims somewhere in there?

It's probably fine as-is, since the parameter has "distributed_claim" in it, but it wouldn't hurt to add it in the description too, something like "...specified by the distributed claim source...".

tvoran (approved):
Don't forget to change the parameter in the PR description from `ms_graph_access_token` to `distributed_claim_access_token`.

path_login.go
```diff
@@ -34,6 +34,10 @@ func pathLogin(b *jwtAuthBackend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: "The signed JWT to validate.",
 			},
+			"distributed_claim_access_token": {
+				Type:        framework.TypeString,
+				Description: "An optional token used to fetch group memberships specified by the distributed claim source in the jwt. This is supported by only supported on Azure/Entra ID",
```
tvoran, line 39:
```suggestion
				Description: "An optional token used to fetch group memberships specified by the distributed claim source in the jwt. This is supported only on Azure/Entra ID",
```

tvoran (approved):

PR #330, opened 2025-04-14
Update CHANGELOG for VAULT-33101

jaireddjawed:
Updates the changelog to account for the change made in this [PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/323/).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## v0.23.1
 
-FEATURES:
+IMPROVEMENTS:
   * Added `normalize_bound_audiences` as a flag that can be enabled to ignore the trailing slash at the end of each bound audience. Bound audiences that differ by only a trailing slash are treated as equivalent. This feature is set to `false` by default.
 
-IMPROVEMENTS:
 * Updated dependencies
   * `github.com/go-jose/go-jose/v3` v3.0.3 -> v3.0.4
   * `github.com/go-jose/go-jose/v4` v4.0.4 -> v4.0.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,42 @@
 ## Unreleased
 
+## v0.23.1
+
+FEATURES:
+  * Added `bound_audience_disregard_trailing_slash` as a flag that can be enabled to ignore the trailing slash at the end of each bound audience. Bound audiences that differ by only a trailing slash are treated as equivalent. This feature is set to `false` by default.
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,43 @@
 ## Unreleased
 
+## v0.23.1
+
+FEATURES:
+  * Added `bound_audience_disregard_trailing_slash` as a flag that can be enabled to ignore the trailing slash at the end of each bound audience. Bound audiences that differ by only a trailing slash are treated as equivalent. This feature is set to `false` by default.
```
tvoran, line 6: (Similar comment here: https://github.com/hashicorp/vault-plugin-auth-jwt/pull/324#discussion_r2029220544)

It looks like this flag is different between [`main`](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/535641386cf8d46dc3891ded44e5bd51fd83bb3b/path_role.go#L123) and the backports? I would really prefer we not change parameters like this when backporting.

jaireddjawed: That name was a leftover from an old version. @benashz wanted me to change it to `normalize_bound_audiences` since `bound_audience_disregard_trailing_slash` is too large of a flag name. But by then, the original PR had already been approved and merged into main. He said that changing it should be fine since there hasn't been any official release.

jaireddjawed: Here is the [PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/331) that changes the flag's name in main.

tvoran: > But by then, the original PR had already been approved and merged into main. He said that changing it should be fine since there hasn't been any official release.

Sure that's fine, but it needs to be changed in `main`, so thank you for opening the PR there. Otherwise the next time a release branch is cut from `main`, we'd end up with the wrong parameter name in the new release.

And next time if you make the changes in `main` first, it should make backporting easier 🙂.

CHANGELOG.md
```diff
@@ -1,11 +1,43 @@
 ## Unreleased
 
+## v0.23.1
+
+FEATURES:
```
digivava, line 5: At least on the Foundations side, we would consider something like this an Improvement, not a Feature. Features generally being larger things that would be worthy of a blog post or some marketing.

It seems like in an older release of this plugin the Improvements section had "Adds abort_on_error parameter to CLI login command to help in non-interactive contexts" which sounds like a similar level of change.

PR #331, opened 2025-04-14
Flag change to main for disregarding the trailing "/"

jaireddjawed:
This PR changes the flag name `bound_audience_disregard_trailing_slash` to `normalize_bound_audiences`. `bound_audience_disregard_trailing_slash` was a new flag introduced from this [PR](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/323). 

The flag was never added into an official release, so @benashz requested that I change it since the original name was too long.


path_role.go
```diff
@@ -463,19 +463,23 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
 		processed := map[string]bool{} // used to prevent duplicate entries
 
 		for _, audience := range role.BoundAudiences {
+			normalizedAudience := audience
+
 			// trim the trailing slash from the audience if it exists
-			audienceWithoutTrailingSlash := strings.TrimRight(audience, "/")
+			if strings.HasSuffix(normalizedAudience, "/") {
+				normalizedAudience = normalizedAudience[:len(normalizedAudience)-1]
+			}
```
tvoran, line 476: Would `strings.TrimSuffix()` work here?

tvoran: Hmm, I don't think that's true. TrimRight is greedy, but TrimSuffix doesn't seem to be: https://go.dev/play/p/bn8_F5x9kIs?v=goprev

digivava: Yeah I am also finding TrimRight to remove all the slashes, and TrimSuffix to just remove the last one.

jaireddjawed: @tvoran Mixed the two names up. Updating the pr now!

path_role.go
```diff
@@ -463,19 +463,23 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
 		processed := map[string]bool{} // used to prevent duplicate entries
 
 		for _, audience := range role.BoundAudiences {
+			normalizedAudience := audience
+
 			// trim the trailing slash from the audience if it exists
-			audienceWithoutTrailingSlash := strings.TrimRight(audience, "/")
+			if strings.HasSuffix(normalizedAudience, "/") {
+				normalizedAudience = strings.TrimSuffix(normalizedAudience, "/")
+			}
```
tvoran, line 476: We probably don't need the `if` check here now, just calling TrimSuffix should suffice?

```suggestion
			normalizedAudience = strings.TrimSuffix(normalizedAudience, "/")
```

path_role.go
```diff
@@ -463,19 +463,21 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
 
 	// disregard the trailing slash (if it exists) on all bound audiences if the flag is set
-	if _, ok := data.GetOk("bound_audience_disregard_trailing_slash"); ok {
+	if _, ok := data.GetOk("normalize_bound_audiences"); ok {
 		boundAudiences := []string{}
 		processed := map[string]bool{} // used to prevent duplicate entries
 
 		for _, audience := range role.BoundAudiences {
+			normalizedAudience := audience
+
```
tvoran, line 472: Sorry one more small nit and I think we're good:

```suggestion
```

path_role.go
```diff
@@ -463,19 +463,21 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
...
 			// trim the trailing slash from the audience if it exists
-			audienceWithoutTrailingSlash := strings.TrimRight(audience, "/")
+			normalizedAudience = strings.TrimSuffix(normalizedAudience, "/")
```
tvoran, line 474:
```suggestion
			normalizedAudience := strings.TrimSuffix(audience, "/")
```

path_role.go
```diff
@@ -463,19 +463,21 @@ func (b *jwtAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical.
 	}
 
 	// disregard the trailing slash (if it exists) on all bound audiences if the flag is set
-	if _, ok := data.GetOk("bound_audience_disregard_trailing_slash"); ok {
+	if _, ok := data.GetOk("normalize_bound_audiences"); ok {
```
tvoran, line 466: I wonder if this should be parsed/type asserted as a bool, and check if it's true or false? Otherwise once someone sets this option, will setting it to `false` actually disable normalization?

path_role.go
```diff
@@ -120,7 +120,7 @@ Defaults to 60 (1 minute) if set to 0 and can be disabled if set to -1.`,
 				Type:        framework.TypeCommaStringSlice,
 				Description: `Comma-separated list of 'aud' claims that are valid for login; any match is sufficient`,
 			},
-			"bound_audience_disregard_trailing_slash": {
+			"normalize_bound_audiences": {
```
tvoran, line 123: Should we return this flag in pathRoleRead()? Otherwise how do we check if it's set?

path_role.go
jaireddjawed: Yep, it should be parsed as a boolean.

PR #365, opened 2026-01-21
Add support for forcefetching groups directly through Graph API

abbyck:
# Overview

This PR enhances the Azure (Entra ID) to support **Direct Group Sourcing**.

This PR addresses two challenges. 
1. Bypassing App-Level Group Limits: A critical limitation in Entra ID is the hard cap of 1,000 groups per application (Service Principal). For organisations that automate group creation this threshold is quickly exhausted. Once this limit is reached, previous mitigations, such as configuring distributed claims, fail because the groups cannot be assigned to the application. Direct Group Sourcing solves this by querying the user's membership directly via the Graph API, decoupling authorisation from the application's group assignment quotas.

2. Currently, the provider relies on the `groups` claim within the `id_token`. This approach has limitations when a user is a member of many groups (Azure omits groups >200) or when the token size exceeds HTTP header limits.

This change allows Vault to retrieve the authenticated user's complete membership information, including transitive group membership, directly from the Microsoft Graph API, bypassing the `id_token` claims entirely when configured.

### What is the change? 

* **New Config Option:** Adds a `fetch_groups` boolean to the `provider_config` for the Azure provider.
* **Graph API Integration:** When `fetch_groups` is enabled, the `FetchGroups` method now utilises the provided `access_token` to query the `https://graph.microsoft.com/v1.0/me/getMemberObjects` endpoint.
* **Fallback Logic:** Maintains backward compatibility. If `fetch_groups` is set to false (the default), the existing distributed claims/token logic is preserved.

### Usage Example

**Configuration:**

```bash
vault write auth/entra-jwt/config \
  oidc_discovery_url="https://login.microsoftonline.com/<tenant-id>/v2.0" \
  oidc_client_id="<client-id>" \
  oidc_client_secret="<client-secret>" \
  provider_config='{"provider": "azure", "fetch_groups": "true"}' \
  default_role= "test-entra"

```

**Login:**

```bash
vault write auth/entra-jwt/login \
   jwt=${ID_TOKEN} \
   distributed_claim_access_token=${DISTRIBUTED_CLAIM_TOKEN}
```

Why is the change needed?
---

- If needed, the feature must be opt-in by setting "fetch_groups": true in the provider configuration.

- The login request to Vault must now include the access_token (passed via the distributed_claim_access_token parameter) alongside the id_token. Without this token, the Graph API call cannot be made, and authentication will fail if the feature is enabled.

## Related Issues/Pull Requests
[x] [PR #329 - allow optional access token to be used to fetch additional groups ](https://github.com/hashicorp/vault-plugin-auth-jwt/pull/329)

### Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/provider_azure.go
+++ b/provider_azure.go
@@ -178,6 +178,9 @@ func (a *AzureProvider) getAzureGroups(groupsURL string, tokenSource oauth2.Toke
 	if err != nil {
 		return nil, fmt.Errorf("unable to get token: %s", err)
 	}
+	if token.AccessToken == "" {
+		return nil, errors.New("access token is empty. Cannot call Microsoft Graph API")
+	}
 	payload := strings.NewReader("{\"securityEnabledOnly\": false}")
 	req, err := http.NewRequest("POST", urlParsed.String(), payload)
 	if err != nil {


provider_azure_test.go
```diff
@@ -158,6 +192,288 @@ func TestLogin_fetchGroups(t *testing.T) {
...
+func TestAzureProvider_FetchGroups_NoTokenSource(t *testing.T) {
+	a := &AzureProvider{
+		ctx: context.Background(),
+		config: AzureProviderConfig{
+			FetchGroups: true,
+		},
+	}
+
+	// Test with nil tokenSource - should fail
+	groups, err := a.getAzureGroups("https://graph.microsoft.com/v1.0/me/getMemberObjects", nil)
```
sachin-chand01: Can we assign this API endpoint to a constant and use it?

provider_azure.go
```diff
@@ -50,6 +63,20 @@ func (a *AzureProvider) SensitiveKeys() []string {
...
+		var err error
+		a.ctx, err = b.createCAContext(b.providerCtx, b.cachedConfig.OIDCDiscoveryCAPEM)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create CA Context: %s", err)
+		}
+		groups, err := a.getAzureGroups(a.buildGraphEndpoint("me/getMemberObjects"), tokenSource)
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch groups from Microsoft Graph API: %s", err)
+		}
+		return groups, nil
```
roh-a: Nit: Should we add a debug log with the response? (similar to Line 75 below)

provider_azure_test.go
abbyck: at present, it is used only once in the tests as a hardcoded

provider_azure_test.go
```diff
@@ -158,6 +192,254 @@ func TestLogin_fetchGroups(t *testing.T) {
...
+			FetchGroups: true,
+		},
+	}
+
+	tokenSource := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: "test.access.token"})
+	groups, err := a.getAzureGroups(server.URL, tokenSource)
+
+	assert.NoError(t, err)
+	assert.Equal(t, expectedGroups, groups)
+}
```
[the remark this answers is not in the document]
abbyck: TestAzureProvider_FetchGroups_WithFetchGroupsEnabled and TestAzureProvider_FetchGroups_ManyGroups now invoke b.fetchGroups with fetch_groups: true instead of calling getAzureGroups directly, so we exercise the full provider path.

The tests assert that we hit POST /v1.0/me/getMemberObjects and validate the expected request payload to ensure the end-to-end FetchGroups flow is covered.

I also added a roundTripFunc helper to redirect graph.microsoft.com traffic to the test server, allowing us to capture and verify the response directly

roh-a (approved):
Looks good to me!

PR #370, opened 2026-03-02
jwt: return 400 when user_claim value is empty string

Android-Login-Analysis:
When the configured user_claim resolves to an empty string in the JWT, createIdentity passes an empty alias name to Vault core, which has a hard constraint that alias names must be non-empty and returns HTTP 500 ("missing name in alias").

Add an explicit check after the string type assertion so the plugin returns a descriptive HTTP 400 instead, consistent with how nil and non-string values are already handled.

The group alias loop in the same function already skips empty strings (line 295); this brings user alias handling in line with that behavior.

# Overview
What is the change? 
Why is the change needed?

# Design of Change

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

## PCI review checklist

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -249,6 +249,9 @@ func (b *jwtAuthBackend) createIdentity(ctx context.Context, allClaims map[strin
 	if !ok {
 		return nil, nil, fmt.Errorf("claim %q could not be converted to string", role.UserClaim)
 	}
+	if userName == "" {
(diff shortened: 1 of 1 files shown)


Android-Login-Analysis:
@heatherezell Thanks for reminding, signed

Android-Login-Analysis:
@heatherezell @hc-github-team-secure-vault-ecosystem Could you please take a look when you get a moment? Thanks

PR #381, opened 2026-04-15
Add kid-based cache for multi-JWKS authentication

suraj-simha:
**Add kid-based cache for multi-JWKS authentication:**

Jira: https://hashicorp.atlassian.net/browse/VAULT-41310

**Summary**
Adds a lightweight kid→KeySet cache to efficiently route JWT validation requests to the correct JWKS endpoint. Instead of iterating through all KeySets, the plugin now performs a fast kid lookup and sends only the matching KeySet to the validator.

**Problem**
When configured with multiple JWKS URLs (e.g., 50 Kubernetes clusters), the validator receives all 50 KeySets and must iterate through each one to find the matching key. This is inefficient and doesn't scale well.

**Solution**
Implement a two-phase kid-based lookup:

Phase 1 (Fast): Check cached kids across all JWKS URLs
Phase 2 (Miss): Fetch kids from remote JWKS endpoints in parallel, then retry
The cache stores only key IDs (kids), not full keys. CAP library continues to handle actual key fetching and JWT verification.

**Testing**
10 new tests covering cache behavior, concurrent access, and two-phase lookup have been added.
All existing tests pass.

# Overview
What is the change? 
Why is the change needed?

# Design of Change

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

## PCI review checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.26.1
 require (
 	github.com/go-jose/go-jose/v3 v3.0.5
 	github.com/go-test/deep v1.1.1
-	github.com/hashicorp/cap v0.12.1-0.20260415172657-9d4dd6dc28b4
+	github.com/hashicorp/cap v0.13.0
 	github.com/hashicorp/errwrap v1.1.0
 	github.com/hashicorp/go-cleanhttp v0.5.2
 	github.com/hashicorp/go-hclog v1.6.3
--- a/go.mod
+++ b/go.mod
@@ -3,13 +3,12 @@ module github.com/hashicorp/vault-plugin-auth-jwt
 go 1.26.1
 
 require (
-	github.com/go-jose/go-jose/v3 v3.0.4
+	github.com/go-jose/go-jose/v3 v3.0.5
 	github.com/go-test/deep v1.1.1
-	github.com/hashicorp/cap v0.12.0
+	github.com/hashicorp/cap v0.12.1-0.20260415172657-9d4dd6dc28b4
 	github.com/hashicorp/errwrap v1.1.0
 	github.com/hashicorp/go-cleanhttp v0.5.2
 	github.com/hashicorp/go-hclog v1.6.3
-	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/go-secure-stdlib/base62 v0.1.2
 	github.com/hashicorp/go-secure-stdlib/httputil v0.1.0
 	github.com/hashicorp/go-secure-stdlib/strutil v0.1.2
@@ -64,6 +63,7 @@ require (
 	github.com/hashicorp/go-kms-wrapping/entropy/v2 v2.0.1 // indirect
 	github.com/hashicorp/go-kms-wrapping/v2 v2.0.18 // indirect
 	github.com/hashicorp/go-metrics v0.5.4 // indirect
+	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/hashicorp/go-plugin v1.6.1 // indirect
 	github.com/hashicorp/go-retryablehttp v0.7.8 // indirect
 	github.com/hashicorp/go-rootcerts v1.0.2 // indirect
--- a/jwks_cache.go
+++ b/jwks_cache.go
@@ -6,7 +6,6 @@ package jwtauth
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -191,17 +190,15 @@ func (b *jwtAuthBackend) findKeySetByKid(ctx context.Context, keyID string) (jwt
 
 	// Cache miss - refresh all caches and retry
 	// Even if refresh has partial failures, we still check if the kid was found
-	refreshErr := b.refreshAllKeySetCaches(ctx, caches)
+	// (Errors are logged in refreshAllKeySetCaches)
+	b.refreshAllKeySetCaches(ctx, caches)
 
 	keySet = findKeySetWithKid(caches, keyID)
 	if keySet != nil {
 		return keySet, nil
 	}
 
-	// Kid still not found after refresh - return error with context
-	if refreshErr != nil {
-		return nil, fmt.Errorf("no key found with kid %s (refresh had errors: %w)", keyID, refreshErr)
-	}
+	// Kid still not found after refresh
 	return nil, fmt.Errorf("no key found with kid %s in any JWKS endpoint", keyID)
 }
 
@@ -223,10 +220,8 @@ func findKeySetWithKid(caches []*JWKSCache, keyID string) jwt.KeySet {
 
 // refreshAllKeySetCaches refreshes kid metadata for all JWKS URLs in parallel.
 // Logs errors but doesn't fail - partial success is acceptable for cache refresh.
-func (b *jwtAuthBackend) refreshAllKeySetCaches(ctx context.Context, caches []*JWKSCache) error {
+func (b *jwtAuthBackend) refreshAllKeySetCaches(ctx context.Context, caches []*JWKSCache) {
 	var wg sync.WaitGroup
-	var mu sync.Mutex
-	var errs []error
 
 	for _, cache := range caches {
 		wg.Add(1)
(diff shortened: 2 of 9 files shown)


jwks_cache.go
```diff
@@ -0,0 +1,347 @@
...
+			ksc.inflightFetch = nil
+			ksc.mu.Unlock()
+
+			// Close channel AFTER updating cache to prevent race conditions
+			close(doneCh)
+		}()
+	}
+
+	inflight := ksc.inflightFetch
+	ksc.mu.Unlock()
```
murali-partha, line 107: It might be better to unlock with a defer statement.

jwks_cache.go
```diff
@@ -0,0 +1,347 @@
...
+	// Even if refresh has partial failures, we still check if the kid was found
+	refreshErr := b.refreshAllKeySetCaches(ctx, caches)
+
+	keySet = findKeySetWithKid(caches, keyID)
+	if keySet != nil {
+		return keySet, nil
+	}
+
+	// Kid still not found after refresh - return error with context
+	if refreshErr != nil {
```
murali-partha: Does it add much value to check this error? Unless we try to refresh again in the same flow, wouldn't it be the same as line 205? We are anyway logging the failures inside `refreshAllKeySetCaches`. Just trying to see if we can avoid the mutex locks inside `refreshAllKeySetCaches` as they are only for catching these errors.

path_config.go
```diff
@@ -374,6 +374,12 @@ func (b *jwtAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Reque
 
 	b.reset()
 
+	// Pre-warm caches for MultiJWKS configuration
+	// This ensures caches are ready by the time users authenticate
+	if len(jwksPairs) > 0 {
+		b.prewarmMultiJWKSCaches(jwksPairs)
```
murali-partha, line 397: Is this activity fast enough to complete in the config creation API call? just curious what would happen if there are a large number of URLs.

murali-partha (commented):
Looks good overall! left some comments

jwks_cache.go
suraj-simha: Using defer here would hold the lock for the entire function (including the wait for the inflight request to complete), which could block other callers, so explicit unlock was intentional here.

jwks_cache.go
suraj-simha: thanks for pointing this! removed it.

path_config.go
suraj-simha: prewarmMultiJWKSCaches() is called synchronously during config write, but it launches goroutines (in warmMultiJWKSCaches()) that fetch key id's in the background. The config API returns immediately without blocking on network fetches. happy to explore improvements if you think we should handle high URL counts differently.

jwks_cache.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// Warm caches in parallel
+	// RefreshKeys has built-in timeout protection via detached context
+	for _, ksc := range caches {
+		go func(cache *JWKSCache, url string) {
+			// Use background context - RefreshKeys handles timeout internally
+			ctx := context.Background()
+			if err := cache.RefreshKeys(ctx); err != nil {
+				b.Logger().Warn("failed to pre-warm JWKS cache", "url", url, "error", err)
+			} else {
```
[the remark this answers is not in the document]
suraj-simha: updated the comments

jwks_cache.go
```diff
@@ -0,0 +1,333 @@
...
+	// If there's not a current inflight request, create one
+	if ksc.inflightFetch == nil {
+		ksc.inflightFetch = &inflight{doneCh: make(chan struct{})}
+
+		// This goroutine has exclusive ownership over the current inflight request.
+		// It releases the resource by nil'ing the field when done.
+		go func() {
+			// Use detached context to prevent one caller's cancellation from affecting others
+			// Add 60s timeout to prevent hanging if JWKS endpoint is unresponsive
+			fetchCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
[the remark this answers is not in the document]
suraj-simha: used ksc.ctx as base context

path_config.go
murali-partha: Cool, feel it is good enough if it is not a blocking call.

murali-partha (dismissed):

go.mod
```diff
@@ -3,13 +3,12 @@ module github.com/hashicorp/vault-plugin-auth-jwt
 go 1.26.1
 
 require (
-	github.com/go-jose/go-jose/v3 v3.0.4
+	github.com/go-jose/go-jose/v3 v3.0.5
 	github.com/go-test/deep v1.1.1
-	github.com/hashicorp/cap v0.12.0
+	github.com/hashicorp/cap v0.12.1-0.20260415172657-9d4dd6dc28b4
```
fairclothjm: Should we request a new tag be cut for CAP?

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+			ksc.inflightFetch.err = err
+			if err == nil {
+				ksc.cachedKids = kids
+			}
+			doneCh := ksc.inflightFetch.doneCh
+			ksc.inflightFetch = nil
+			ksc.mu.Unlock()
+
+			// Close channel AFTER updating cache to prevent race conditions
+			close(doneCh)
```
fairclothjm, line 102: should we close this channel before we Unlock? It seems like there may be a chance of a race condition and multiple inflight requests. If we want to be extra cautious, we could also check if it was already closed.

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+// Logs errors but doesn't fail - partial success is acceptable for cache refresh.
+func (b *jwtAuthBackend) refreshAllKeySetCaches(ctx context.Context, caches []*JWKSCache) {
+	var wg sync.WaitGroup
+
+	for _, cache := range caches {
+		wg.Add(1)
+		go func(ksc *JWKSCache) {
+			defer wg.Done()
+
+			if err := ksc.RefreshKeys(ctx); err != nil {
```
fairclothjm, line 234: What happens if all caches fail? Does the caller need to have visibility or can we ignore that?

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+				b.Logger().Warn("error refreshing JWKS kid cache",
```
fairclothjm, line 235: Will this log be too noisy? How often do we expect failures?

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+	case <-inflight.doneCh:
+		return inflight.err
+	}
+}
+
+// fetchKids performs a lightweight HTTP fetch to get only the key IDs from the JWKS endpoint.
+// The inflight mechanism prevents thundering herd on concurrent requests.
+func (ksc *JWKSCache) fetchKids(ctx context.Context) ([]string, error) {
+	// Extract HTTP client from context (configured with CA certs via createCAContext)
+	client := http.DefaultClient
```
fairclothjm, line 122: Does this need a timeout? or is the 60s context timeout sufficient?

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read JWKS response: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("JWKS endpoint returned status %d: %s", resp.StatusCode, string(body))
```
fairclothjm: Do we need to return the full body response? Is that going to be large?

jwks_cache.go
```diff
@@ -0,0 +1,334 @@
...
+	b.l.RLock()
+	caches := b.jwksCaches
+	b.l.RUnlock()
+
+	// Warm caches in parallel
+	// RefreshKeys has built-in timeout protection via detached context
+	for _, ksc := range caches {
+		go func(cache *JWKSCache, url string) {
+			// Use background context - RefreshKeys handles timeout internally
+			ctx := context.Background()
```
fairclothjm: Should this be the provider context instead?

fairclothjm (commented):
Looking great! Left some comments/questions

go.mod
suraj-simha: yes, will request and update mod before merging.

jwks_cache.go
suraj-simha: Good point! The current order is safe. Setting `inflightFetch = nil` before unlocking is the protection which prevents new goroutines from joining while goroutines already waiting have their own saved copy of the pointer. Both orders work I think.

jwks_cache.go
suraj-simha: thanks for pointing this, updated to return an error if all refreshes fail.

jwks_cache.go
suraj-simha: we do not expect much failures, infact we expect the kid to be in cache most of the times as the keys don't change often.

jwks_cache.go
suraj-simha: 60s of context timeout alone should be sufficient here.

jwks_cache.go
suraj-simha: thanks, full body is not necessary. Removed it.

jwks_cache.go
suraj-simha: yep, updated it to use the provider context.

fairclothjm (dismissed):
LGTM!

yaminisridaran (approved):
Looks good to me

murali-partha (approved):
LGTM!

PR #403, opened 2026-05-27
Adding Okta Provider for fetching all the groups from admin okta API

suraj-simha:
Jira : https://hashicorp.atlassian.net/browse/VAULT-43952

## Overview
Adds Okta provider to JWT/OIDC auth plugin with group fetching that handles Okta's silent group truncation in ID tokens.

## Problem
Okta truncates groups in ID tokens when users exceed a threshold (default: 100), leading to incomplete group assignments in Vault.

## Solution
Implements two-phase group fetching:
- **Fast path**: Uses token groups when `groups < groups_cap` (no API call)
- **Slow path**: Fetches from Okta Admin API when truncation detected (`groups >= groups_cap`)

## Key Features
- **`fetch_groups` flag**: Enable/disable API fallback (default: `false` for backward compatibility)
- **Auto-derive `org_url`**: Automatically extracts from `oidc_discovery_url` if not provided
- **Truncation detection**: Automatically detects when groups are truncated at boundary based on `groups_cap`
- **Pagination support**: Handles users with large number of groups
- **Groups filtering**: Regex-based filtering with `groups_filter`
- **Flexible user ID**: Configurable via `user_id_claim`, falls back to role's `user_claim`

## Configuration Example
```hcl
provider_config = {
  provider       = "okta"
  org_url        = "https://example.okta.com"  # Optional if oidc_discovery_url set
  api_token      = "00abc...xyz"               # Required only when fetch_groups=true
  fetch_groups   = true                        # Enable API fallback
  groups_cap     = 100                         # Truncation threshold
  groups_filter  = "^vault-"                   # Optional regex filter
  user_id_claim  = "email"                     # Optional, defaults to role's user_claim
}
```

## Files Changed
- **New**: `provider_okta.go` (450 lines) - Provider implementation
- **New**: `provider_okta_test.go` (950 lines) - 39 comprehensive test cases
- **Modified**: `provider_config.go` - Register Okta provider

## Backward Compatibility
This addition is fully backward compatible - `fetch_groups` defaults to `false` (token-only mode)

# Overview
What is the change? 
Why is the change needed?

# Design of Change

# Contributor Checklist
[My Docs PR Link](link)
[ ] Backwards compatible

## PCI review checklist

### DIFF
--- a/provider_okta.go
+++ b/provider_okta.go
@@ -1,0 +1,285 @@
+package jwtauth
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"regexp"
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"golang.org/x/oauth2"
+)
+
+const (
+	// defaultOktaGroupsCap is Okta's default ID-token groups-claim
+	// truncation threshold. When a user belongs to more groups than
+	// this, Okta drops groups silently rather than aggregating
+	// distributed claims. Configurable per Okta authorization server;
+	// override via provider_config.groups_cap.
+	defaultOktaGroupsCap = 100
+
+	// oktaUserGroupsPathFmt is the Okta admin API path that returns
+	// the groups a user belongs to. The {user} segment accepts an
+	// Okta user id, login (email/UPN), or unique login shortname.
+	oktaUserGroupsPathFmt = "/api/v1/users/%s/groups"
+)
+
+// OktaProvider returns the full set of Okta groups a user belongs to
+// when Okta has truncated the id-token's groups claim. It calls Okta's
+// admin endpoint
+//
+//	GET /api/v1/users/{user}/groups
+//
+// authenticated with a configured Okta API token (SSWS), following
+// RFC 5988 Link-header pagination.
+//
+// This endpoint is admin-only and cannot be reached with an end-user
+// OAuth token regardless of okta.* scopes. The provider therefore
+// requires an Okta API token bound to a user with permission to read
+// users and read groups (a least-privilege custom admin role granting
+// "View users and their details" and "View groups and their details"
+// is sufficient). The token is supplied via provider_config.api_token
+// and is masked from config reads.
+type OktaProvider struct {
+	ctx    context.Context
+	config OktaProviderConfig
+
+	// groupsFilter is the compiled form of OktaProviderConfig.GroupsFilter.
+	// Nil when no filter is configured.
+	groupsFilter *regexp.Regexp
+}
+
+// OktaProviderConfig is decoded from jwtConfig.ProviderConfig during
+// Initialize.
+type OktaProviderConfig struct {
+	// OrgURL is the Okta org base URL, e.g. https://example.okta.com.
+	// Must use https.
+	OrgURL string `mapstructure:"org_url"`
+
+	// APIToken is an Okta API token (SSWS) used to authenticate the
+	// server-side groups lookup. Required.
+	APIToken string `mapstructure:"api_token"`
+
+	// UserIDClaim names the claim in the OIDC token whose value
+	// identifies the user to Okta. The value must be one of: the
+	// user's Okta id, login (email/UPN), or unique login shortname.
+	// Optional; when empty, the role's user_claim is used.
+	UserIDClaim string `mapstructure:"user_id_claim"`
+
+	// GroupsCap is the Okta id-token groups truncation threshold. When
+	// the groups claim is present with length >= GroupsCap, the
+	// provider treats it as truncated and re-fetches the full list
+	// via the admin API. Defaults to 100.
+	GroupsCap int `mapstructure:"groups_cap"`
+
+	// GroupsFilter is an optional Go regular expression. When
+	// non-empty, only groups whose name matches the pattern are
+	// returned to Vault. Applied identically to both the id-token
+	// claim path and the API fallback path so behavior is consistent
+	// regardless of whether the user crossed GroupsCap. Empty (the
+	// default) disables filtering; every group Okta returns passes
+	// through.
+	GroupsFilter string `mapstructure:"groups_filter"`
+}
+
+// Initialize validates and stores provider configuration. Satisfies
+// the CustomProvider interface.
+func (o *OktaProvider) Initialize(_ context.Context, jc *jwtConfig) error {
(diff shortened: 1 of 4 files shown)


provider_okta.go
```diff
@@ -0,0 +1,341 @@
+// Copyright IBM Corp. 2018, 2025
```
sachin-chand01: Can we update this to 2026 ?

suraj-simha: done, thanks!

provider_okta.go
```diff
@@ -0,0 +1,341 @@
...
+// When fetch_groups is disabled, the provider uses only the groups from
+// the ID token, which may be incomplete if the user belongs to more groups
+// than the configured groups_cap threshold.
+type OktaProvider struct {
+	ctx    context.Context
+	config OktaProviderConfig
+
+	// groupsFilter is the compiled form of OktaProviderConfig.GroupsFilter.
+	// Nil when no filter is configured.
+	groupsFilter *regexp.Regexp
```
roh-a, line 67: Is there a reason why groupsFilter is stored here instead of the the `OktaProviderConfig` struct?

provider_okta.go
```diff
@@ -0,0 +1,341 @@
...
+func (o *OktaProvider) SensitiveKeys() []string {
+	return []string{"api_token"}
+}
+
+// FetchGroups implements GroupsFetcher. Returns the id-token groups
+// claim when it's clearly untruncated; otherwise calls the Okta
+// admin API using the configured SSWS token. When groups_filter is
+// set, both paths return only the subset of group names matching the
+// configured regex.
+func (o *OktaProvider) FetchGroups(_ context.Context, b *jwtAuthBackend, allClaims map[string]interface{}, role *jwtRole, _ oauth2.TokenSource) (interface{}, error) {
```
roh-a: Here we are not using the `Context`. Should we be using the Context for the API call during login operation so that context cancellations can be preserved.

provider_okta.go
```diff
@@ -0,0 +1,341 @@
...
+	client := http.DefaultClient
+	if c, ok := o.ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
+		client = c
+	}
+
+	next := strings.TrimRight(o.config.OrgURL, "/") + fmt.Sprintf(oktaUserGroupsPathFmt, url.PathEscape(userID))
+
+	var all []interface{}
+	for next != "" {
+		req, err := http.NewRequest("GET", next, nil)
```
roh-a: The pagination logic follows 'next' links indefinitely without a maximum page limit. If Okta's API returns a circular reference or the Link header is manipulated, this could cause an infinite loop that hangs authentication requests. 

Does it make sense to add a maximum page count to protect against infinite loops.

provider_okta.go
```diff
@@ -0,0 +1,341 @@
...
+			return nil, fmt.Errorf("unable to call Okta API: %w", err)
+		}
+		body, readErr := io.ReadAll(res.Body)
+		res.Body.Close()
+		if readErr != nil {
+			return nil, fmt.Errorf("failed to read Okta API response: %w", readErr)
+		}
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("okta api returned %d: %s", res.StatusCode, string(body))
+		}
```
roh-a, line 319: Could the error message here expose/return the SSWS token? Let's validate this for all the places we are returning the error response. 
Should we sanitize the error responses before returning it?

provider_okta.go
```diff
@@ -0,0 +1,341 @@
...
+// /api/v1/users/{userID}/groups, following RFC 5988 Link-header
+// pagination. Authenticated with the configured admin API token.
+func (o *OktaProvider) getOktaGroups(userID string) ([]interface{}, error) {
+	client := http.DefaultClient
+	if c, ok := o.ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
+		client = c
+	}
+
+	next := strings.TrimRight(o.config.OrgURL, "/") + fmt.Sprintf(oktaUserGroupsPathFmt, url.PathEscape(userID))
+
```
roh-a, line 295: here `userID` parameter is used directly without validation. Do we need to validate the userID for invalid characters? 
Do we need character escapes here?

roh-a (commented):
Looking good! Leaving some comments.

provider_okta.go
suraj-simha: *regexp.Regexp cannot be decoded by [mapstructure] (it only handles primitive types and simple structs), so it can't live in OktaProviderConfig. It needs to be compiled from the raw GroupsFilter string after decoding. Added a comment on the field explaining this.

provider_okta.go
suraj-simha: updated. Thanks!

provider_okta.go
suraj-simha: good catch. There is no documented maximum number of pages in okta, so added a max page limit of 200 now.
can update it if we think this is not sufficient. Each page contains 100 groups.

provider_okta.go
suraj-simha: The body included in the non-200 response error is Okta's own JSON error response. The SSWS token is only placed in the Authorization request header and never appears in any error path.

provider_okta.go
suraj-simha: resolveUserID rejects empty strings with an explicit error. The user ID is passed through [url.PathEscape(userID)] before being placed in the URL, which percent-encodes all characters that are not valid in a URL path segment

roh-a (dismissed):
Looks good to me!

provider_okta.go
```diff
@@ -0,0 +1,356 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid org_url: %w", err)
+		}
+		if u.Scheme != "https" {
+			return errors.New("org_url must use https")
+		}
+		if u.Host == "" {
+			return errors.New("org_url must include a host")
+		}
+	}
```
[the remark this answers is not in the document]
sachin-chand01: @suraj-simha  Could we perform a quick manual test to validate this?

sachin-chand01 (dismissed):
Left one minor comment,  otherwise  LGTM

provider_okta.go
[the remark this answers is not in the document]
suraj-simha: thanks, updated to take `scheme://host` only.

sachin-chand01 (approved):

roh-a (approved):

